PR #5, opened 2020-05-01
Integrate Mujoco

Zhehui-Huang:


algorithms/appo/model_utils.py
```diff
@@ -257,6 +257,35 @@ def forward(self, obs_dict):
         return x
 
 
+class MujocoEncoder(EncoderBase):
```
alex-petrenko, line 260: Don't introduce a new class. Just add another if in encode_subtype check in the existing MLP encoder. Because I see that this encoder is exactly the same as MLP encoder.

.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <settings>
```
alex-petrenko, line 2: Please don't commit these files. Use git status command, it shows which files you're adding to the commit. Or you can add these to the .gitignore

envs/mujoco/README.md
```diff
@@ -0,0 +1,33 @@
...
+- Put the Mujoco key under $HOME/.mujoco/
+
+- Install extra environments
+
+```
+cd ~/Downloads
+git clone https://github.com/openai/mujoco-py.git
+bash -c "cd mujoco-py && conda activate sample-factory && python setup.py install"
+sudo apt-get update -y
+sudo apt-get install -y patchelf
```
alex-petrenko, line 31: Why do we need mujoco-py from github? Isn't it better to just install from Pip?

envs/mujoco/mujoco_params.py
```diff
@@ -0,0 +1,21 @@
...
+        env_frameskip=1,
+        nonlinearity='tanh'
+    )
+
+
+# noinspection PyUnusedLocal
+def add_mujoco_env_args(env, parser):
+    p = parser
+
+    p.add_argument('--mujoco_discretize_actions', default=-1, type=int, help='Discretize actions into N bins for each individual action. Default (-1) means no discretization')
```
alex-petrenko, line 19: Please remove these, they are specific to quadrotors

PR #7, opened 2020-05-03
Integrate mujoco v1

Zhehui-Huang:


algorithms/appo/model_utils.py
```diff
@@ -274,6 +282,8 @@ def create_standard_encoder(cfg, obs_space, timing):
         encoder = ResnetEncoder(cfg, obs_space, timing)
     elif cfg.encoder_type == 'mlp':
         encoder = MlpEncoder(cfg, obs_space, timing)
+    elif cfg.encoder_type == 'mujoco':
```
alex-petrenko, line 285: Again, why do we need the MujocoEncoder? MlpEncoder will do!

envs/mujoco/test_env.py
```diff
@@ -0,0 +1,14 @@
+import sys
+
+import gym
+
+
+def main():
```
alex-petrenko, line 6: Do we really need this file in the repo? I recommend either writing the test (like Python unit-test) or removing this altogether

runner/runs/mujoco_halfcheetah_grid_search.py
```diff
@@ -0,0 +1,60 @@
+from runner.run_description import RunDescription, Experiment, ParamGrid
+
+_params = ParamGrid([
```
alex-petrenko, line 3: This does not search over any seeds. To actually try multiple seeds you should add e.g. 'seed': [1,2,3,4] to the ParamGrid

algorithms/appo/model_utils.py
Zhehui-Huang: My mistake, I forgot to delete this.

envs/mujoco/test_env.py
Zhehui-Huang: No, I will delete this.

runner/runs/mujoco_halfcheetah_grid_search.py
Zhehui-Huang: This file just like an example, and I make our parameter to be the same as rllib ppo, besides ppo_clip_ratio, since I found to set it as 4 is better than set it as 0.2.

Zhehui-Huang: Any way, i will do some changes to make it looks like grid search

PR #12, opened 2020-05-21
fix one minor error under envs/doom/doom_gym.py

Zhehui-Huang:


envs/doom/doom_gym.py
```diff
@@ -393,7 +393,10 @@ def _process_game_step(self, state, done, info):
             observation = self._black_screen()
 
             # when done=True Doom does not allow us to call get_info, so we provide info from the last frame
-            info.update(self._prev_info)
+            try:
+                info.update(self._prev_info)
+            except:
+                log.info('_prev_info is None')
```
alex-petrenko, line 399: It is not a good fix for several reasons.
Most importantly, this update() call is supposed to provide valid info when done=True
If you fix it like this the info will remain None

Please do something like:

if info is None:
  info = self._prev_info
else:
  info.update(self._prev_info)

PR #13, opened 2020-05-30
Removed fast_queue and replaced it with faster-fifo pip package

tushartk:


setup.py
```diff
@@ -1,27 +0,0 @@
-from Cython.Build import cythonize
```
alex-petrenko, line 1: Please don't delete the setup.py file. It allows to install the entire sample factory as Pip package. Just remove everything related to the fast queue from it.

PR #17, opened 2020-06-09
Deleted unnecessary test case files

tushartk:


algorithms/appo/tests/test_appo_utils.py
```diff
@@ -1,23 +1,7 @@
...
-                self.assertEqual(cores, [i % cpu_count])
-            elif i == 40:
-                self.assertEqual(cores, [0, 1, 2, 3, 4])
-            elif i == 41:
-                self.assertEqual(cores, [5, 6, 7, 8, 9])
-            elif i == 42:
-                self.assertEqual(cores, [10, 11, 12, 13, 14])
-            elif i == 43:
-                self.assertEqual(cores, [15, 16, 17, 18, 19])
+        pass
```
alex-petrenko, line 6: Don't delete this. I think this was a legit test.

utils/tests/test_utils.py
```diff
@@ -1,10 +1,23 @@
...
-
-from utils.network import is_udp_port_available
+from utils.utils import cores_for_worker_process
 
 
 class TestNetwork(TestCase):
-    def test_udp(self):
-        is_udp_port_available(50301)
+    def test_cpu_affinity(self):
+        num_workers = 44
```
alex-petrenko, line 7: This does not make sense. Why would you remove the is_udp_port_available line and add the rest of the code instead?
I think everything was okay and you shouldn't have changed it

algorithms/appo/tests/test_appo_utils.py
tushartk: I've shifted this to test_utils

utils/tests/test_utils.py
tushartk: The cores_for_worker_process method was in the utils folder and hence I shifted it here

PR #18, opened 2020-06-15
Refactored tests to make them work

tushartk:
1. Please verify the assertions in the `test_runner.py` scripts
2. I've also commented out `multi_env_performance` tests from `test_envs.py` for now, as they are now initialized in a different way. I will be working on those next.
3. Have also deleted `create_multi_env` function from `env_utils.py`. For now only `EnvCriticalError` Exception class remains, which could perhaps be shifted out to `doom_gym.py` as that's the only place where it is used.

### DIFF
--- a/envs/env_utils.py
+++ b/envs/env_utils.py
@@ -1,29 +1,2 @@
-from utils.utils import log
-
-
-def create_multi_env(num_envs, num_workers, make_env_func, stats_episodes):
-    """
-    Create a vectorized env for single- and multi-agent case. This is only required for synchronous algorithms
-    such as PPO and A2C. APPO uses a different mechanism with separate worker processes.
-    """
-
-    tmp_env = make_env_func(None)
-    is_multiagent = hasattr(tmp_env, 'num_agents') and tmp_env.num_agents > 1
-
-    if is_multiagent:
-        assert num_envs % tmp_env.num_agents == 0
-        log.debug('Num envs %d agents %d', num_envs, tmp_env.num_agents)
-        num_envs = num_envs // tmp_env.num_agents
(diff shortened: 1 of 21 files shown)


envs/env_utils.py
```diff
@@ -1,29 +1,2 @@
...
-    else:
-        from algorithms.utils.multi_env import MultiEnv
-        multi_env = MultiEnv(num_envs, num_workers, make_env_func, stats_episodes)
-
-    tmp_env.close()
-
-    return multi_env
-
-
 class EnvCriticalError(Exception):
```
alex-petrenko, line 1: I think you can leave it here, it's fine. Don't move it to Doom

envs/tests/test_envs.py
```diff
@@ -55,36 +55,37 @@ def test_env_performance(make_env, env_type, verbose=False):
...
-                log.info('Collected %d frames of experience...', frames)
-                next_print += print_step
-
-    fps = total_num_frames / t.experience
-    log.debug('%s performance:', env_type)
-    log.debug('Took %.3f sec to collect %d frames in parallel, %.1f FPS', t.experience, total_num_frames, fps)
-    log.debug('Timing: %s', t)
-
-    multi_env.close()
+# def test_multi_env_performance(make_env, env_type, num_envs, num_workers, total_num_frames=1000):
```
alex-petrenko, line 61: This test relied on the old class MultiEnv
Now we can rewrite it using DummySampler class
Or just remove this test completely

tushartk: Sure, probably thought I would rewrite it

envs/env_utils.py
tushartk: Perfect

PR #19, opened 2020-06-18
Added travis CI configuration for autorunning the tests

tushartk:
- Edited the test to check for GPU in test_model.py
- The reason a new environment.yml was created was to install CPU only Pytorch and its dependencies. Hence that has been kept in the .travis folder to read from.


.travis/environment.yml
```diff
@@ -0,0 +1,140 @@
+name: sample-factory
```
alex-petrenko, line 1: Why do we have two versions of environment.yml?
If they are different, can you put the difference in a patch file (like the output git diff generates), and patch the main file in this script? This way it will be a lot more apparent.
Otherwise, code duplication is really bad, especially of this sort. We will change something in main .yml and forget to change it here.

algorithms/appo/tests/test_model.py
```diff
@@ -25,7 +25,9 @@ def test_forward_pass(self):
         torch.backends.cudnn.benchmark = True
 
         actor_critic = create_actor_critic(cfg, env.observation_space, env.action_space)
-        device = torch.device('cuda')
+        device_type = 'gpu' if torch.cuda.is_available() else 'cpu'
```
alex-petrenko, line 28: Can you please instead make this a parameter of the test function, and then skip the GPU test with @unittest.skipUnless when CUDA is not available

envs/doom/multiplayer/tests/test_vizdoom_multiplayer.py
```diff
@@ -51,7 +51,7 @@ def test_doom_multiagent(self):
         self.doom_multiagent(self.make_standard_dm, worker_index=0)
 
     def test_doom_multiagent_parallel(self):
-        num_workers = 16
+        num_workers = 1
```
alex-petrenko, line 54: The whole point of this test is running VizDoom in parallel. If this does not work in Travis (why wouldn't it by the way??)
then you should create two tests and skip one in Travis somehow. I do not approve this change, unless you explain why it is needed.

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+  - source "$HOME/miniconda/etc/profile.d/conda.sh"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  - conda info -a
+  - conda env create -q -f ./.travis/environment.yml
+  - conda activate sample-factory
+
+script:
+  - python -m unittest
```
alex-petrenko, line 27: python -m unittest will crash on Travis for the reason I described on Slack (see test_example comments)

.travis/environment.yml
tushartk: The problem is with Travis. Travis would require CPU only versions of torch

.travis.yml
tushartk: It did not. If you notice the build has passed.

alex-petrenko: That's weird. Most likely the CPU-only version of PyTorch makes a difference (see the github issue for details https://github.com/pytorch/pytorch/issues/33248

Well, I guess it makes our lives easier.
Do all tests pass on your machine too?

tushartk: Yes

algorithms/appo/tests/test_model.py
tushartk: So do you mean to convert this into 2 different tests? and skip the GPU one if it's not available right?

.travis/environment.yml
tushartk: Patch file created

algorithms/appo/tests/test_model.py
alex-petrenko: yes

.travis.yml
alex-petrenko: Might be something related to UDP ports given to different game instances

tushartk: Yup, I think it is the UDP port issue as this Exception is being thrown from there - raise RuntimeError('Critical error: worker stuck on initialization. Abort!')

alex-petrenko:
Well this does not tell a whole lot

пт, 19 июн. 2020 г. в 00:12, Tushar K <notifications@github.com>:

.travis.yml
tushartk: It was a really weird bug that fixed itself. Now it's running fine.

alex-petrenko: Looks like it does not depend on the number of processes. Rather, it's a inconsistent issue. Like, these multiplayer games only sometimes (rarely) just fail to start and are therefore killed by the parent process. When you run more processes in parallel there is just a higher chance of it reproducing.

I have seen an error like this a few times when I was running duel & deathmatch experiments for the paper. It is not a big issue in SampleFactory because it has the reset mechanism for workers on the initial startup. I.e. when we detect a freeze like this one, we can just restart the whole worker. In this test, there is no such mechanism of course.

It usually happens very rarely, but perhaps something on Travis increases the probability of this issue. E.g. the fact that the CPU there is very slow.

The ideal course of action would be try to identify if this is an issue on our side vs in VizDoom. If it's on our side, we should fix the issue that leads to freezes. If it's in VizDoom and it's not obvious how to fix this, we should embed a workaround for this into the multiplayer game wrapper itself, such that it won't rely on the external SampleFactory mechanism (need to think how to do it properly). The wrapper solution can also work if we fail to identify the root cause for these freezes.
Also, does this test fail on your machine?

One (probably unrelated) thing that just came to mind, we should replace regular queues with faster-fifo. This should increase perfrormance basically for free because of lower latency. This sounds like a perfect task for you actually!
You might need to rewrite the code there a bit to get rid of the JoinableQueue dependency. I think a regular Queue should work just fine

tushartk: Yes, it fails on my machine. That was my system's log. This is the other exception from Vizdoom which gets thrown too:-
```
VizDoom game.init() threw an exception ViZDoomUnexpectedExitException('Controlled ViZDoom instance exited unexpectedly.'). Terminate process...
```
What I notice is that around `num_workers = 8` these failures start off.

algorithms/appo/tests/test_model.py
```diff
@@ -11,8 +12,10 @@
 
 
 class TestModel(TestCase):
-    def test_forward_pass(self):
-        env_name = 'atari_breakout'
+
+    @staticmethod
+    def forward_pass(device_type):
+        env_name = 'doom_benchmark'
```
alex-petrenko, line 18: I deliberatly edited this test to not require doom, so it can be run without installing any additional envs. This merge will fail because we both edited the same function. Please merge master into your branch (using my changes, e.g. atari and different obs resolution) and re-submit the pull request.

algorithms/appo/tests/test_model.py
```diff
@@ -43,3 +46,10 @@ def test_forward_pass(self):
                 log.debug('Progress %d/%d', i, n)
 
         log.debug('Timing: %s', timing)
+
+    def test_forward_pass_cpu(self):
+        self.forward_pass('cpu')
+
+    @unittest.skipUnless(torch.cuda.is_available(), "This test requires a GPU")
```
alex-petrenko, line 53: This is good. Exactly what I meant.

Please use single quotes everywhere.

PR #20, opened 2020-06-23
Added code coverage with badges for the same and for license

tushartk:
- Added code coverage with the badge
- Added license badge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,10 @@ install:
   - patch environment.yml ./.travis/patchfile.patch
   - conda env create -q -f environment.yml
   - conda activate sample-factory
+  - pip install coverage
 
 script:
-  - python -m unittest
+  - coverage run -m unittest
+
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -23,6 +23,10 @@ install:
...
   - conda env create -q -f environment.yml
   - conda activate sample-factory
+  - pip install coverage
 
 script:
-  - python -m unittest
+  - coverage run -m unittest
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
alex-petrenko, line 32: Is this a standard practice :D
Download some shell script from the internet and execute it, every time? I don't care because it's Travis and not our own buildbot, but this looks weird to me. Generally not a good practice to do that.

tushartk: Yup, this is standard practice. On their website itself, this is mentioned.

PR #21, opened 2020-06-24
Added environment variable for setting timeout in travis

tushartk:


envs/doom/multiplayer/doom_multiagent.py
```diff
@@ -120,9 +124,13 @@ def _ensure_initialized(self):
                 self.game.add_game_args(f'-record {demo_path}')
         else:
             # Join existing game.
+            try:
+                vizdoom_env_timeout = int(os.environ['TRAVIS_VIZDOOM_ENV_TIMEOUT'])
```
alex-petrenko, line 128: I think this code will be better if you query the environment variable outside of if.
Is there a reason for this code to be duplicated? Usually it is always better to avoid code duplication, DRY https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#DRY_vs_WET_solutions

envs/doom/multiplayer/doom_multiagent.py
```diff
@@ -97,9 +97,13 @@ def _ensure_initialized(self):
                 '+sv_nojump 1',  # Disables jumping.
                 '+sv_nofreelook 1',  # Disables free look with a mouse (only keyboard).
                 '+sv_noexit 1',  # Prevents players from exiting the level in deathmatch before timelimit is hit.
-                f'+viz_respawn_delay {self.respawn_delay}',  # Sets delay between respanws (in seconds).
-                '+viz_connect_timeout 4',  # In seconds
+                f'+viz_respawn_delay {self.respawn_delay}'  # Sets delay between respanws (in seconds).
```
alex-petrenko, line 100: It's good to have a comma for the last element in the list, and Python allows this.
Minor stylistic thing since I'm requesting another change anyway

envs/doom/multiplayer/doom_multiagent.py
```diff
@@ -8,6 +8,12 @@
 DEFAULT_UDP_PORT = int(os.environ.get('DOOM_DEFAULT_UDP_PORT', 40300))
 log.info('Default UDP port is %r', DEFAULT_UDP_PORT)
 
+# This try except block is increase the env timeout connection flag in travis
+try:
+    vizdoom_env_timeout = int(os.environ['TRAVIS_VIZDOOM_ENV_TIMEOUT'])
```
alex-petrenko, line 13: Since you moved this to the global scope, it's better to follow the convention and use the FULL_CAPS name for it. But it's okay. Actually I like it in the global scope

PR #39, opened 2020-07-26
Add commit hash to config file

tushartk:


utils/utils.py
```diff
@@ -328,3 +329,10 @@ def cfg_file(cfg):
 
 def done_filename(cfg):
     return join(experiment_dir(cfg=cfg), 'done')
+
+
+def get_git_commit_hash():
+    path_to_project = os.path.dirname(os.path.realpath(__file__))
+    return subprocess.check_output(['git', 'rev-parse', 'HEAD'],
+                                   cwd=path_to_project,
+                                   timeout=0.1).strip().decode('ascii')
```
alex-petrenko, line 338: If you're using a timeout here, it's a good idea to catch the exception thrown by timeout arg. Also, 0.1 is a pretty low value, can potentially fail in some oversubscribed system or in the cloud where the filesystem tends to be a little slow at startup. Change it to 5 seconds or something.

Also, consider that git might just not be available if this is some sort of Docker container, or Slurm instance, or Cubernetes instance. It's a good idea to catch the exception here and pass return some fallback string (e.g. 'unknown'). See e.g. `exception subprocess.CalledProcessError` in the subprocess docs. https://docs.python.org/3/library/subprocess.html#subprocess.SubprocessError

PR #44, opened 2020-08-14
Changed instances of entropy_coeff to exploration_coeff

tushartk:


benchmarks/rlpyt/wall_time_doom_ppo.py
```diff
@@ -213,7 +213,7 @@ def build_and_train(game="doom_benchmark", run_ID=0, cuda_idx=None, n_parallel=-
     algo = PPO(
         learning_rate=0.0001,
         value_loss_coeff=0.5,
-        entropy_loss_coeff=0.003,
+        exploration_loss_coeff=0.003,
```
alex-petrenko, line 216: This is not a parameter of our algorithm. Look at the file path, this is the RLPYT benchmark! Pls fix

tushartk: Fixed this

tushartk:
Yes I searched for the keyword everywhere and fixed it

PR #56, opened 2020-09-23
Add two benchmark runner scripts for quadrotor env

Zhehui-Huang:
Alex, I have added two benchmark runner scripts for quadrotor env. So, it would be more convenient for Sumeet and Tushar to do the sanity check. And I have tested these two scripts on my own local machine. There should not be any bugs for us to directly use these two scripts. Please check if you have time.

### DIFF
--- a/envs/quadrotors/quadrotor_params.py
+++ b/envs/quadrotors/quadrotor_params.py
@@ -23,4 +23,4 @@ def add_quadrotors_env_args(env, parser):
     p.add_argument('--quads_collision_reward', default=None, type=float, help='Override default value for quadcol_bin reward')
     p.add_argument('--quads_settle_reward', default=None, type=float, help='Override default value for quadsettle reward')
     p.add_argument('--quads_dist_between_goals', default=0.3, type=float, help='Under circular configuration scenarios, it should be the radius of the circle of goals')
-    p.add_argument('--quads_mode', default='circular_config', type=str, choices=['circular_config', 'same_goal'], help='Choose which scanerio to run')
+    p.add_argument('--quads_mode', default='sanity_check', type=str, choices=['circular_config', 'same_goal', 'sanity_check'], help='Choose which scanerio to run')
--- a/runner/runs/quad_multi_benchmark.py
+++ b/runner/runs/quad_multi_benchmark.py
@@ -1,0 +1,34 @@
+from runner.run_description import RunDescription, Experiment, ParamGrid
+
+_params = ParamGrid([
+    ('seed', [0000, 1111, 2222, 3333, 4444, 5555, 6666, 7777]),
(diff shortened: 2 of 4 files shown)


runner/runs/quad_multi_benchmark.py
```diff
@@ -0,0 +1,34 @@
+from runner.run_description import RunDescription, Experiment, ParamGrid
+
+_params = ParamGrid([
+    ('seed', [0000, 1111, 2222, 3333, 4444, 5555, 6666, 7777]),
+    ('quads_num_agents', [4]),
```
alex-petrenko, line 5: What's the point of specifying these parameters here in a gridsearch?
They could be just regular command line parameters.
If you want this for readability, consider just splitting the CMD string parts and += some special parameters in the end.

Because currently these parameters are here, they will be made a part of the filenames in tensorboard etc. which will result in very long and nonsensical names.

envs/quadrotors/quadrotor_params.py
```diff
@@ -23,4 +23,4 @@ def add_quadrotors_env_args(env, parser):
     p.add_argument('--quads_collision_reward', default=None, type=float, help='Override default value for quadcol_bin reward')
     p.add_argument('--quads_settle_reward', default=None, type=float, help='Override default value for quadsettle reward')
     p.add_argument('--quads_dist_between_goals', default=0.3, type=float, help='Under circular configuration scenarios, it should be the radius of the circle of goals')
-    p.add_argument('--quads_mode', default='circular_config', type=str, choices=['circular_config', 'same_goal'], help='Choose which scanerio to run')
+    p.add_argument('--quads_mode', default='sanity_check', type=str, choices=['circular_config', 'same_goal', 'sanity_check'], help='Choose which scanerio to run')
```
alex-petrenko, line 26: *scenario

alex-petrenko (approved):

envs/quadrotors/quadrotor_params.py
Zhehui-Huang: Could you please explain what's the meaning of *scenario? Is this mean that you would recommend me to change the name from quads_mode to quads_scenario?

alex-petrenko: No, there was a typo in your help string, "scanerio".
I already fixed it in master.

runner/runs/quad_multi_benchmark.py
Zhehui-Huang: Sure, I will modify it.

envs/quadrotors/quadrotor_params.py
Zhehui-Huang: I see, thanks.

PR #58, opened 2020-09-25
Mean Embeddings Encoder for MultiDrone 

SumeetBatra:
Added a new boolean param "extend_obs" in quadrotor_params.py to specify whether you want to feed in the extended observation space or the original obs_space. Default=False

(PS neither the new param nor the encoder will run until the gym_art PR has been merged)


runner/runs/quad_single_gridsearch.py
```diff
@@ -17,7 +17,8 @@
 
 _experiment = Experiment(
     'quads_gridsearch',
-    'python -m run_algorithm --env=quadrotor_single --train_for_env_steps=100000 --algo=APPO --gamma=0.99 --use_rnn=False --num_workers=36 --num_envs_per_worker=2 --num_policies=1 --rollout=32 --recurrence=32 --benchmark=False --with_pbt=False --ppo_clip_ratio=0.05',
+    'python -m run_algorithm --env=quadrotor_single --train_for_env_steps=100000 --algo=APPO --gamma=0.99 --use_rnn=False --num_workers=8 --num_envs_per_worker=2 --num_policies=1 --rollout=32 --recurrence=32 --benchmark=False --with_pbt=False --ppo_clip_ratio=0.05',
```
Zhehui-Huang, line 20: I think we don't have to change this file, how about remove this change?

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
+import torch
+from torch import nn
+
+from algorithms.appo.model_utils import get_obs_shape, nonlinearity, create_standard_encoder, EncoderBase, \
+    register_custom_encoder
+from algorithms.utils.pytorch_utils import calc_num_elements
+from utils.utils import log
```
Zhehui-Huang, line 7: We don't need `from utils.utils import log`, right?

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class QuadMultiMeanEncoder(EncoderBase):
+    '''Mean embedding encoder based on the DeepRL for Swarms Paper'''
+    def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6, neighbor_hidden_size=32):
```
Zhehui-Huang, line 12: Again, we don't need `obs_space` here, since we the main idea to pass obs_space here is to get observation shape, (take a look at doom_model.py). But here, since we already have self_obs_dim and neighbor_obs_dim, we don't need obs_space anymore.

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+        super().__init__(cfg, timing)
+        self.self_obs_dim = self_obs_dim
+        self.neighbor_obs_dim = neighbor_obs_dim
+        self.neighbor_hidden_size = neighbor_hidden_size
+        self.device = torch.device('cpu' if cfg.device == 'cpu' else 'cuda')
```
Zhehui-Huang, line 17: Do we need self.device here?

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+        self.neighbor_encoder = nn.Sequential(
+            nn.Linear(self.neighbor_obs_dim, self.neighbor_hidden_size),
+            nonlinearity(cfg),
+        )
+        self.self_encoder_out_size = calc_num_elements(self.self_encoder, (self.self_obs_dim,))
+        self.neighbor_encoder_out_size = calc_num_elements(self.neighbor_encoder, (self.neighbor_obs_dim,))
+
+        # Feed forward self obs and neighbor obs after concatenation
+        self.feed_forward = nn.Linear(self.self_encoder_out_size + self.neighbor_encoder_out_size, 64)
+        self.feed_forward_out_size = calc_num_elements(self.feed_forward, (self.self_encoder_out_size + self.neighbor_encoder_out_size,))
```
Zhehui-Huang, line 37: We don't need Line 37, `self.feed_forward_out_size = calc_num_elements(self.feed_forward, (self.self_encoder_out_size + self.neighbor_encoder_out_size,))`, right?

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+        self.init_fc_blocks(self.feed_forward_out_size)
```
Zhehui-Huang, line 39: And we also don't need this line

algorithms/appo/learner.py
```diff
@@ -963,6 +963,7 @@ def initialize(self, timing):
                 self.device = torch.device('cuda', index=0)
             else:
                 self.device = torch.device('cpu')
+
```
Zhehui-Huang, line 966: How about delete this change?

runner/runs/quad_single_gridsearch.py
Zhehui-Huang, line 20: How about resume to the previous version? Since I don't think change `num_workers`  from 36 to 8 make sense.

Zhehui-Huang (commented):
@SumeetBatra , I found that there are 11 commits, it seems too many. Just suggestions, how about make it only one commit? Since it could make the main codebase more clear.

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+
+    def forward(self, obs_dict):
+        obs = obs_dict['obs']
+        # TODO: Don't hardcode obs_self and obs_neighbors (pass a labeled dict??)
+        obs_self, obs_neighbors = obs[:, :18], obs[:, 18:]
```
Zhehui-Huang, line 46: Instead of using 18 here, how about use `self.self_obs_dim`?

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+from torch import nn
+
+from algorithms.appo.model_utils import get_obs_shape, nonlinearity, create_standard_encoder, EncoderBase, \
+    register_custom_encoder
+from algorithms.utils.pytorch_utils import calc_num_elements
+from utils.utils import log
+
+
+class QuadMultiMeanEncoder(EncoderBase):
+    '''Mean embedding encoder based on the DeepRL for Swarms Paper'''
```
alex-petrenko, line 11: Please adhere to the code style in the rest of the codebase. We use single quotes ' for all strings, and triple double quotes for docstrings.
I have no particular preference for either, just for consistency.

envs/quadrotors/quad_multi_model.py
alex-petrenko, line 17: This is not going to work with multiple GPUs, 'cuda' means GPU #0, in a multi-GPU system you need to pass an index as well, like `cuda:0`

Also, Zhehui is right, I don't think you need self.device at all.
And if you need it in forward function, you can always use obs.device

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+        # encode the neighboring drone's observations
+        self.neighbor_encoder = nn.Sequential(
+            nn.Linear(self.neighbor_obs_dim, self.neighbor_hidden_size),
+            nonlinearity(cfg),
+        )
+        self.self_encoder_out_size = calc_num_elements(self.self_encoder, (self.self_obs_dim,))
+        self.neighbor_encoder_out_size = calc_num_elements(self.neighbor_encoder, (self.neighbor_obs_dim,))
+
+        # Feed forward self obs and neighbor obs after concatenation
+        self.feed_forward = nn.Linear(self.self_encoder_out_size + self.neighbor_encoder_out_size, 64)
```
alex-petrenko, line 36: I'd replace 64 with cfg.hidden_size to make it somewhat configurable.

envs/quadrotors/quad_multi_model.py
alex-petrenko, line 37: Again, you know the output size here, no need to call calc_num_elements.

envs/quadrotors/quad_multi_model.py
alex-petrenko, line 39: I think in your case it is redundant to call init_fc_blocks
Looks like this class ignores the `encoder_extra_fc_layers` parameter anyway.

But if you remove it, don't forget to set `self.encoder_out_size` to the correct value, because it's a part of the interface.

envs/quadrotors/quad_multi_model.py
alex-petrenko: Agree with Zhehui, we already have a parameter for that.

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+    def forward(self, obs_dict):
+        obs = obs_dict['obs']
+        # TODO: Don't hardcode obs_self and obs_neighbors (pass a labeled dict??)
+        obs_self, obs_neighbors = obs[:, :18], obs[:, 18:]
+        self_embed = self.self_encoder(obs_self)
+        neighbor_embeds = []
+        # TODO: Find a way to fold neighbor_obs (dim1) onto batch_size dim (dim 0) so we don't have to for-loop
+        for i in range(0, obs_neighbors.shape[1], self.neighbor_obs_dim):
```
alex-petrenko, line 50: We should try to do it without for-loop. This will be very slow. You're doing this on the GPU, remember, and GPUs are great at doing a few computations on a lot of data, but terrible at doing many computations on small pieces of data.

This can be vectorized with torch.split and torch.reshape.
1) separate neighbor obs along axis=1 with torch.split
2) reshape them so batch dimension (dim=0) is batch_size x num_neighbors and obs shape dimension (dim=1) is 6 (xyz_vxyz)
3) apply forward pass
4) reshape them back into single vectors of size 32
5) calculate the mean

envs/quadrotors/quad_utils.py
```diff
@@ -48,6 +49,13 @@ def make_quadrotor_env_single(cfg, **kwargs):
 
 def make_quadrotor_env_multi(cfg, **kwargs):
     from gym_art.quadrotor_multi.quadrotor_multi import QuadrotorEnvMulti
+    global CUSTOM_ENCODER_INITIALIZED
+
+    # TODO: Make this less hacky
```
alex-petrenko, line 54: Yeah, not a fan of this stuff myself. If you have a better idea, let me know :D
I think CUSTOM_ENCODER_INITIALIZED is not even required here, you can just check in register_models if the encoder is already registered, and skip if it is.

alex-petrenko (changes requested):

envs/quadrotors/quad_multi_model.py
```diff
@@ -33,24 +31,21 @@ def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6,
...
+        self.feed_forward = nn.Linear(self.self_encoder_out_size + self.neighbor_encoder_out_size, cfg.hidden_size)
 
-        self.init_fc_blocks(self.feed_forward_out_size)
+        self.init_fc_blocks(cfg.hidden_size)
 
 
 
     def forward(self, obs_dict):
         obs = obs_dict['obs']
+        batch_size = obs.shape[0]
```
Zhehui-Huang, line 42: I think we don't need this line.

envs/quadrotors/quad_multi_model.py
```diff
@@ -33,24 +31,21 @@ def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6,
...
-        obs_self, obs_neighbors = obs[:, :18], obs[:, 18:]
+        obs_self, obs_neighbors = obs[:, :self.self_obs_dim], obs[:, self.self_obs_dim:]
         self_embed = self.self_encoder(obs_self)
-        neighbor_embeds = []
-        # TODO: Find a way to fold neighbor_obs (dim1) onto batch_size dim (dim 0) so we don't have to for-loop
-        for i in range(0, obs_neighbors.shape[1], self.neighbor_obs_dim):
-            embed = self.neighbor_encoder(obs_neighbors[:, i:i+self.neighbor_obs_dim])
-            neighbor_embeds.append(embed)
-        mean_embed = torch.mean(torch.stack(neighbor_embeds), 0)
+        obs_neighbors = torch.stack(torch.split(obs_neighbors, 6, dim=1))
```
Zhehui-Huang, line 46: How about change 6 to self.neighbor_obs_dim

Zhehui-Huang (changes requested):

algorithms/appo/learner.py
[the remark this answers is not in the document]
SumeetBatra: I don't remember writing this, might've been a previous commit from another person that was later undone. I'll remove it anyways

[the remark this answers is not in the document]
SumeetBatra: Removing this results in an assertion error, so I believe it is being used somewhere (unless the assert causing the error is also unneeded).

envs/quadrotors/quad_utils.py
SumeetBatra: What was the reasoning behind registering models and having an ENCODER_REGISTRY?

alex-petrenko: This allows SampleFactory to be used as an external dependency.
So one can just install it as a pip package instead of hacking directly in the repo. And define their custom architectures to be used by SampleFactory.
See examples/train_custom_env_custom_model.py for details.

I think encoder registry makes sense (just a set of all supported encoders), just like environment registry.
What's redundant here is the CUSTOM_ENCODER_INITIALIZED variable.

envs/quadrotors/quad_multi_model.py
```diff
@@ -0,0 +1,53 @@
...
+    def forward(self, obs_dict):
+        obs = obs_dict['obs']
+        obs_self, obs_neighbors = obs[:, :self.self_obs_dim], obs[:, self.self_obs_dim:]
+        self_embed = self.self_encoder(obs_self)
+        obs_neighbors = torch.stack(torch.split(obs_neighbors, self.neighbor_obs_dim, dim=1))
+        neighbor_embeds = self.neighbor_encoder(obs_neighbors)
+        mean_embed = torch.mean(neighbor_embeds, 0)
+        embeddings = torch.cat((self_embed, mean_embed), dim=1)
+        out = self.feed_forward(embeddings)
+        return out
```
alex-petrenko, line 49:
```
    def forward(self, obs_dict):
        obs = obs_dict['obs']
        obs_self, obs_neighbors = obs[:, :self.self_obs_dim], obs[:, self.self_obs_dim:]
        self_embed = self.self_encoder(obs_self)

        # relative xyz and vxyz for the entire minibatch (batch dimension is batch_size * num_neighbors)
        obs_neighbors = obs_neighbors.reshape(-1, self.neighbor_obs_dim)
        neighbor_embeds = self.neighbor_encoder(obs_neighbors)

        batch_size = obs_self.shape[0]
        neighbor_embeds = neighbor_embeds.reshape(batch_size, -1, self.neighbor_hidden_size)
        mean_embed = torch.mean(neighbor_embeds, dim=1)

        embeddings = torch.cat((self_embed, mean_embed), dim=1)
        out = self.feed_forward(embeddings)
        return out
```

Here's a version without split/stack

alex-petrenko (changes requested):

algorithms/appo/learner.py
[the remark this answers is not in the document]
Zhehui-Huang: Yeah, Alex is right, I just meant the whitespace.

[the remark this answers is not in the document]
SumeetBatra: lol I think this mightve been left over from when I was logging stuff to see how the hell SF works. I'll be careful of stray newlines in the future

envs/quadrotors/quad_utils.py
SumeetBatra: I don't know how to get rid of this variable since without it, I get assertion errors saying that the encoder is already in the registry. I refactored the code to make it the same as how the encoder is added in doom_utils.py. Do you have any suggestions for other ways of doing this?

alex-petrenko (approved):
I changed a couple of minor things, I think we're good now!
Let's fire up some experiments!

Zhehui-Huang (commented):
Awesome.

PR #60, opened 2020-10-01
Added enabling of numba with a flag in sample factory

tushartk:


envs/quadrotors/tests/test_quads.py
```diff
@@ -34,20 +57,13 @@ def test_quad_multi_env(self):
...
-        timing = Timing()
-        with timing.timeit('step'):
-            for i in range(n_frames):
-                obs, r, d, info = env.step([env.action_space.sample() for _ in range(env.num_agents)])
-
-        log.debug('Time %s, FPS %.1f', timing, n_frames * env.num_agents / timing.step)
-        env.close()
+    @unittest.skipUnless(quadrotors_available(), 'quadrotor env package is not installed')
+    @unittest.skipUnless(numba_available(), 'Numba is not installed')
+    def test_quad_multi_env_with_numba(self):
```
alex-petrenko, line 64: This won't run in Travis, right? Because quadrotors_available check will fail.

I'm fine with keeping this here, although it'd be better if the tests that we already have in quadrotor env were executed in Travis

alex-petrenko: Maybe consider adding a Travis config to Artem's repo also?

alex-petrenko (approved):
Great!
Let's start using this right away!

envs/quadrotors/tests/test_quads.py
tushartk: I'll ask him to enable and perhaps enable it for that

tushartk: Oh I just saw that Travis does not work for private repositories

PR #63, opened 2020-10-02
Support num_agents=1 under quad_multi env

Zhehui-Huang:
For supporting num_agents=1 under quad_multi env, we don't have to use MultiAgentWrapper, since we already wrap the obs, rew, done, info in multi_quad env, and the simplest way is setting variable is_multiagent=1 under quad_multi env.


alex-petrenko (approved):
I guess this is fine too.

What I had in mind was defining is_multiagent in the multi quad env, so that later wrappers can just query this flag.

But this works too.

Zhehui-Huang:
I see, I will modify this in the next version / PR.

PR #64, opened 2020-10-03
Add histogram method, modify support num_agent=1

Zhehui-Huang:
Alex, there are two commits, the first one is about modifying support num_agent=1, I changed it according to your comments: set is_multiagent as a parameter in multi-quad env. The second is about adding the histogram method. I also modified runner scripts.


envs/quadrotors/quad_multi_model.py
```diff
@@ -52,7 +52,49 @@ def forward(self, obs_dict):
...
+            nn.Linear(self.self_obs_dim, fc_encoder_layer),
+            nonlinearity(cfg),
+            nn.Linear(fc_encoder_layer, fc_encoder_layer),
+            nonlinearity(cfg)
+        )
+
+        self.histogram_encoder = nn.Sequential(
+            nn.Linear(self.histogram_bins, self.histogram_hidden_size),
+            nonlinearity(cfg),
+        )
```
SumeetBatra, line 74: @alex-petrenko Correct me if I'm wrong, but I don't think we have separate encoders for self_observations and the histogram(?) The histogram embedding should be concatenated with self_observations and this should be fed into the policy_network that spits out actions.

PR #66, opened 2020-10-26
Spectral Norm

SumeetBatra:


algorithms/appo/appo.py
```diff
@@ -115,6 +115,8 @@ def add_cli_args(cls, parser):
         )
 
         p.add_argument('--max_grad_norm', default=4.0, type=float, help='Max L2 norm of the gradient vector')
+        #TODO: Right now this only applies to custom encoders. Make sure generic policies also factor in this arg
+        p.add_argument('--use_spectral_norm', default=False, type=str2bool, help='Use spectral normalization to smoothen the gradients and stabilize training')
```
alex-petrenko, line 119: I think the comment should mention the fact that it is only supported for fully-connected layers now, just for complete clarity.

algorithms/appo/model_utils.py
```diff
@@ -270,12 +271,20 @@ def __init__(self, cfg, obs_space, timing):
...
+                    spectral_norm(nn.Linear(fc_encoder_layer, fc_encoder_layer)),
+                    nonlinearity(cfg),
+                ]
+            else:
+                encoder_layers = [
+                    nn.Linear(obs_shape.obs[0], fc_encoder_layer),
+                    nonlinearity(cfg),
+                    nn.Linear(fc_encoder_layer, fc_encoder_layer),
+                    nonlinearity(cfg),
+                ]
```
alex-petrenko, line 287: This definitely works, but quite a lot of code duplication where we can easily avoid it.

How about defining a function `fc_layer` somewhere in model_utils.py which will return either wrapped or unwrapped nn.Linear depending on the config?
Then we can use it everywhere Linear layers are used (except for the outputs of the policies, I think there FC layers should be without the spectral norm)

envs/quadrotors/quad_multi_model.py
```diff
@@ -13,25 +14,42 @@ def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6,
...
+                nonlinearity(cfg)
+            )
         # encode the neighboring drone's observations
-        self.neighbor_encoder = nn.Sequential(
-            nn.Linear(self.neighbor_obs_dim, self.neighbor_hidden_size),
-            nonlinearity(cfg),
-        )
+        if self.use_spectral_norm:
+            self.neighbor_encoder = nn.Sequential(
+                spectral_norm(nn.Linear(self.neighbor_obs_dim, self.neighbor_hidden_size)),
```
alex-petrenko, line 38: Unrelated to spectral norm, but do you think it makes sense to do 2 fc layers before feature aggregation?
Intuitively, just a single linear transformation + tanh might not be enough to get the best features.

I'd definitely try 2 fc layers before mean() and concat()

alex-petrenko (changes requested):

Zhehui-Huang (commented):
I agree with the comments of Alex.

alex-petrenko (approved):
Awesome!

envs/quadrotors/quad_multi_model.py
```diff
@@ -13,25 +14,30 @@ def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6,
...
+            fc_layer(self.neighbor_hidden_size, self.neighbor_hidden_size, spec_norm=self.use_spectral_norm),
+            nonlinearity(cfg)
         )
+
         self.self_encoder_out_size = calc_num_elements(self.self_encoder, (self.self_obs_dim,))
         self.neighbor_encoder_out_size = calc_num_elements(self.neighbor_encoder, (self.neighbor_obs_dim,))
 
         # Feed forward self obs and neighbor obs after concatenation
-        self.feed_forward = nn.Linear(self.self_encoder_out_size + self.neighbor_encoder_out_size, cfg.hidden_size)
+        self.feed_forward = fc_layer(self.encoder_out_size + self.neighbor_encoder_out_size, cfg.hidden_size, spec_norm=self.use_spectral_norm)
```
Zhehui-Huang, line 40: We can not change `self.self_encoder_out_size` to `self.encoder_out_size`, right? Since we have never defined `self.encoder_out_size`. I would fix it and create a PR.

Zhehui-Huang (commented):
I think there is a bug in this commit.

PR #67, opened 2020-10-29
Modify num_agents=1, fix a minor bug

Zhehui-Huang:
1. Modify num_agents=1, more details, you can take a look at the comment from Alex, https://github.com/alex-petrenko/sample-factory/pull/63. I just found that We haven't merged this commit, since for last time, I create a PR including 2 commits, this one and the histogram method.
2. I fixed a bug introduced by our latest PR, more details, you can take a look at the comment from me, https://github.com/alex-petrenko/sample-factory/pull/66
I would also create a corresponding PR for gym_art, @alex-petrenko, could you please also take a look?

### DIFF
--- a/envs/quadrotors/quad_multi_model.py
+++ b/envs/quadrotors/quad_multi_model.py
@@ -37,7 +37,7 @@ class QuadMultiMeanEncoder(EncoderBase):
         self.neighbor_encoder_out_size = calc_num_elements(self.neighbor_encoder, (self.neighbor_obs_dim,))
 
         # Feed forward self obs and neighbor obs after concatenation
-        self.feed_forward = fc_layer(self.encoder_out_size + self.neighbor_encoder_out_size, cfg.hidden_size, spec_norm=self.use_spectral_norm)
+        self.feed_forward = fc_layer(self.self_encoder_out_size + self.neighbor_encoder_out_size, cfg.hidden_size, spec_norm=self.use_spectral_norm)
 
         self.init_fc_blocks(cfg.hidden_size)
 
--- a/envs/quadrotors/wrappers/reward_shaping.py
+++ b/envs/quadrotors/wrappers/reward_shaping.py
@@ -11,7 +11,7 @@ DEFAULT_QUAD_REWARD_SHAPING = dict(
 
 
 class QuadsRewardShapingWrapper(gym.Wrapper, RewardShapingInterface):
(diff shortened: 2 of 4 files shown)


envs/quadrotors/quad_multi_model.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6,
         self.neighbor_encoder_out_size = calc_num_elements(self.neighbor_encoder, (self.neighbor_obs_dim,))
 
         # Feed forward self obs and neighbor obs after concatenation
-        self.feed_forward = fc_layer(self.encoder_out_size + self.neighbor_encoder_out_size, cfg.hidden_size, spec_norm=self.use_spectral_norm)
+        self.feed_forward = fc_layer(self.self_encoder_out_size + self.neighbor_encoder_out_size, cfg.hidden_size, spec_norm=self.use_spectral_norm)
```
alex-petrenko, line 40: I trust you that this works. Are we still defining the encoder_out_size? I think it is used in some other code

envs/quadrotors/wrappers/reward_shaping.py
```diff
@@ -49,7 +48,7 @@ def step(self, action):
             env_reward_shaping[key] = weight
 
         obs, rewards, dones, infos = self.env.step(action)
-        if self.is_multiagent:
+        if hasattr(self.env, 'is_multiagent') and self.env.is_multiagent is True:
```
alex-petrenko, line 51: As I said, it is better to cache the result of this whole expression (e.g. save it to the class variable), since hasattr queries are rather slow.
Also, instead of `self.env.is_multiagent is True` just write `self.env.is_multiagent`

I will approve for now but please change this in the following commit.

alex-petrenko (approved):

envs/quadrotors/quad_multi_model.py
Zhehui-Huang: We defined encoder_out_size in class EncoderBase, and initialize it to -1. However, we have never defined or update the value of encoder_out_size in this file (quad_multi_model.py). Therefore, we should definitely avoid to use encoder_out_size here, unless we set self.encoder_out_size =  self.self_encoder_out_size

alex-petrenko:
Yes that's what I meant.
I'm just wondering if code which uses this class doesn't use
encoder_out_size

I trust you that you tested this and it works

PR #71, opened 2020-11-06
Arguably the most important 3 letters of the 21st Century 

SumeetBatra:
They call me the Einstein of robotics. 

Jokes aside, this will let us see weighted reward components in tensorboard :)


alex-petrenko (approved):
Awesome

PR #80, opened 2021-01-19
Use local info of neighbors, Global Camera, Simulate gravity of obstacles

Zhehui-Huang:


envs/quadrotors/quadrotor_params.py
```diff
@@ -36,6 +36,10 @@ def add_quadrotors_env_args(env, parser):
     p.add_argument('--quads_obstacle_num', default=0, type=int, help='Choose the number of obstacle(s)')
     p.add_argument('--quads_obstacle_type', default='sphere', type=str, choices=['sphere', 'cube'], help='Choose the type of obstacle(s)')
     p.add_argument('--quads_obstacle_size', default=0.0, type=float, help='Choose the size of obstacle(s)')
+    p.add_argument('--quads_obstacle_traj', default='gravity', type=str, choices=['gravity', 'electron'],  help='Choose the type of force to use')
+    p.add_argument('--quads_neighbor_info_mode', default='local', type=str, choices=['local', 'global'], help='Choose the type of neighbor information to use')
+    p.add_argument('--quads_ratio_use_neighbor_obs', default=0.0, type=float, help='0: Not use the info of other drones, 1: Use info of all of other drones, n in (0, 1): Use the info of n * (num_agents-1) closest drones')
+
```
alex-petrenko, line 42: Okay, I can't say I'm a fan of this.
I think it would be better to use a single integer value here instead of this floating point ratio.
This will not necessarily scale with the swarm size. I.e. if we find that 4 neighbors works best, when we increase the swarm size from 8 to 40 this value should remain 4, instead of becoming 20.

So, I think it's not the ratio that we should keep between experiments but the absolute value of the number of neighbors.

alex-petrenko (approved):
I approve and merge for now, but please address my comments

PR #81, opened 2021-01-19
extend obs space to include goal pos info 

SumeetBatra:


envs/quadrotors/quad_multi_model.py
```diff
@@ -11,11 +11,15 @@ class QuadMultiMeanEncoder(EncoderBase):
     def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6, neighbor_hidden_size=32, obstacle_obs_dim=6, obstacle_hidden_size=32):
         super().__init__(cfg, timing)
         self.self_obs_dim = self_obs_dim
-        self.neighbor_obs_dim = neighbor_obs_dim
+        if cfg.obs_space == 'extend+': self.neighbor_obs_dim = 9  # include goal pos info
+        else: self.neighbor_obs_dim = neighbor_obs_dim
```
alex-petrenko, line 15: Do we actually ever pass non-default neighbor_obs_dim parameter to this function?
I think it only causes extra confusion now.

We should just set self.neighbor_obs_size based on `cfg.obs_space`

envs/quadrotors/quad_multi_model.py
```diff
@@ -56,7 +60,7 @@ def __init__(self, cfg, obs_space, timing, self_obs_dim=18, neighbor_obs_dim=6,
         # this is followed by another fully connected layer in the action parameterization, so we add a nonlinearity here
         self.feed_forward = nn.Sequential(
             fc_layer(total_encoder_out_size, cfg.hidden_size, spec_norm=self.use_spectral_norm),
-            nonlinearity(cfg),
+            nn.Tanh(),
```
alex-petrenko, line 63: Fair. I remember this. I think we can do it.

envs/quadrotors/quadrotor_params.py
```diff
@@ -30,15 +30,14 @@ def add_quadrotors_env_args(env, parser):
     p.add_argument('--quads_settle_range_meters', default=1.0, type=float, help='Radius of the sphere around the goal with velocity penalty to help quadrotors stop and settle at the goal')
     p.add_argument('--quads_spacing_coeff', default=0.0, type=float, help='Override default coefficient for spacing penalty between drones ')
 
-    p.add_argument('--extend_obs', default=False, type=str2bool, help='Drones receive relative pos and relative vel info from all other drones')
+    p.add_argument('--obs_space', default='default', type=str, choices=['default', 'extend', 'extend+'], help='Choose what kind of obs to send to encoder. Extend = rel. pos/vel. Extend+ = rel. pos.vel and goal pos')
```
alex-petrenko, line 33: See comments in gym_art about the parameter names

envs/quadrotors/quadrotor_params.py
```diff
@@ -30,15 +30,14 @@ def add_quadrotors_env_args(env, parser):
...
     p.add_argument('--quads_obstacle_num', default=0, type=int, help='Choose the number of obstacle(s)')
     p.add_argument('--quads_obstacle_type', default='sphere', type=str, choices=['sphere', 'cube'], help='Choose the type of obstacle(s)')
     p.add_argument('--quads_obstacle_size', default=0.0, type=float, help='Choose the size of obstacle(s)')
     p.add_argument('--quads_obstacle_traj', default='gravity', type=str, choices=['gravity', 'electron'],  help='Choose the type of force to use')
-    p.add_argument('--quads_neighbor_info_mode', default='local', type=str, choices=['local', 'global'], help='Choose the type of neighbor information to use')
-    p.add_argument('--quads_ratio_use_neighbor_obs', default=0.0, type=float, help='0: Not use the info of other drones, 1: Use info of all of other drones, n in (0, 1): Use the info of n * (num_agents-1) closest drones')
+    p.add_argument('--quads_local_obs', default=-1, type=int, help='Number of neighbors to consider. -1=all neighbors. 0=blind agents, 0<n<num_agents-1 = nonzero number of agents')
```
alex-petrenko, line 40: Okay, this is good.
Do we ever check that quads_local_obs is < num_agents?

alex-petrenko (changes requested):

envs/quadrotors/quadrotor_params.py
SumeetBatra: Yes in quadrotor_multi

envs/quadrotors/quad_multi_model.py
SumeetBatra: Default is meant to be equivalent to extend_obs=False from before

envs/quadrotors/quad_multi_model.py
```diff
@@ -11,11 +11,23 @@ class QuadMultiMeanEncoder(EncoderBase):
...
-        self.num_use_neighbor_obs = int(cfg.quads_ratio_use_neighbor_obs * (cfg.quads_num_agents - 1))
+        if cfg.quads_local_obs == -1:
+            self.num_use_neighbor_obs = cfg.quads_num_agents - 1
+        else:
+            self.num_use_neighbor_obs = cfg.quads_local_obs
+
+        if self.neighbor_obs_type == 'pos_vel_goals':
+            self.neighbor_obs_dim = 9  # include goal pos info
+        elif self.neighbor_obs_type == 'pos_vel':
+            self.neighbor_obs_dim = neighbor_obs_dim
```
alex-petrenko, line 26: I think this should be explicitly `= 6` and we should just remove `neighbor_obs_dim` function parameter because we actually never pass it. It's just an extra moving part.

Also, I think `else` should check for explicitly `None` and then another `else` with `throw NotImplemented()`

envs/quadrotors/quad_multi_model.py
```diff
@@ -11,11 +11,23 @@ class QuadMultiMeanEncoder(EncoderBase):
...
-        self.neighbor_obs_dim = neighbor_obs_dim
+        self.neighbor_obs_type = cfg.neighbor_obs_type
         self.neighbor_hidden_size = neighbor_hidden_size
         self.use_spectral_norm = cfg.use_spectral_norm
         self.obstacle_mode = cfg.quads_obstacle_mode
-        self.num_use_neighbor_obs = int(cfg.quads_ratio_use_neighbor_obs * (cfg.quads_num_agents - 1))
+        if cfg.quads_local_obs == -1:
+            self.num_use_neighbor_obs = cfg.quads_num_agents - 1
+        else:
+            self.num_use_neighbor_obs = cfg.quads_local_obs
```
alex-petrenko, line 21: Maybe assert that this is a correct value.

alex-petrenko (approved):
Good. Should work!

envs/quadrotors/quad_multi_model.py
SumeetBatra: This is done in quadrotor_multi.py line 36

PR #85, opened 2021-01-27
Add CPC|A since it is super useful

erikwijmans:
Paper: https://arxiv.org/abs/1811.06407


alex-petrenko (approved):
Not going to pretend I understood everything in the code on the first pass, but seems very reasonable!

I still stand that adding rewards to observation embeddings would've been nice! (unless you tried it and it didn't work :)

erikwijmans:
> I still stand that adding rewards to observation embeddings would've been nice! (unless you tried it and it didn't work :)

I didn't give that a try, it would be worth trying tho.  I don't think they did the the original paper, but I do think they did it here: http://proceedings.mlr.press/v119/guo20g/guo20g-supp.pdf

PR #87, opened 2021-01-28
Working attention mechanism

SumeetBatra:


Zhehui-Huang (changes requested):
I have tested the attention encoder, and double-check the logic and shape of each layers carefully. I think the attention encoder here is basically the same as the paper Crowd-Robot Interaction, except for the final layer.
Besides, I also added the local map, which name Mi in the paper to our project. And I have already created PRs for that under Sumeet's repo.

alex-petrenko:
This seems to be failing tests... Can you guys please re-merge master into this branch and try again?

alex-petrenko (approved):

PR #88, opened 2021-01-30
Setup actor workers to make team spirit increase over time

erikwijmans:


algorithms/appo/actor_worker.py
```diff
@@ -169,6 +170,13 @@ def record_env_step(self, reward, done, info, traj_buffer_idx, rollout_step):
             self.last_episode_true_reward = info.get('true_reward', self.last_episode_reward)
             self.last_episode_extra_stats = info.get('episode_extra_stats', dict())
 
+    def process_policy_steps(self, policy_steps):
+        if self.cfg.increase_team_spirit and self.last_policy_steps != policy_steps:
+            self.last_policy_steps = float(policy_steps)
```
alex-petrenko, line 175: I think this `last_policy_steps` should be per-policy. Won't matter in most cases, but if for some reason different policies collect experience at different rate, this is going to affect the calculation.

alex-petrenko (approved):

algorithms/appo/actor_worker.py
erikwijmans: The VectorEnvRunner uses the actor state's policy to call this.  I did `!=` so that even if policy steps is less, it'll still be updated.  I think this works.

PR #99, opened 2021-02-20
Change the observation space of obstacles

Zhehui-Huang:


alex-petrenko (changes requested):
Can we just do the obstacles observation change without the extra runner scripts?
I feel like there are too many already...
We should keep the obstacles one in this PR of course!

alex-petrenko (approved):

PR #100, opened 2021-02-25
Plots script

Zhehui-Huang:


alex-petrenko (approved):
I think we should move it to a separate folder, such as plots/quadrotors

PR #102, opened 2021-03-02
fix: catch exception 'report queue is full' in actor_worker

lanpartis:
this bug terminates actor workers unexpectedly

### DIFF
--- a/algorithms/appo/actor_worker.py
+++ b/algorithms/appo/actor_worker.py
@@ -870,6 +870,8 @@ class ActorWorker:
                     self.report_queue.put(dict(critical_error=self.worker_idx))
                 except KeyboardInterrupt:
                     self.terminate = True
+                except Full:
+                    log.warning('Could not report training stats, the report queue is full!')
                 except:
                     log.exception('Unknown exception in rollout worker')
                     self.terminate = True


alex-petrenko (approved):
Thank you for fixing this!

While I think that it is useful to catch the exception, I think there must be a more fundamental problem for the queue overflow. Perhaps the main loop is stuck or is too slow to keep up. Or the total size of the stats is very high, multiplied by the number of workers really causes the overflow. In this case, it is a good idea to increase the max size of the circular buffer in the queue, or reduce the size of the summaries.

The max size of the queue can be changed in the faster-fifo queue constructor

alex-petrenko:
Are you running this with a custom environment or with one of the standard envs? Could you please send the command line for your experiment?

alex-petrenko:
@lanpartis 
I did the handling a bit more carefully in this commit : https://github.com/alex-petrenko/sample-factory/commit/dc872adf32e22b021cdb34aeff153014fc4efc5e

Please let me know if you see any error messages. If we can find a fundamental problem that leads to this exception, that'd be great.

lanpartis:
Thank you for the advice! I think increasing the capacity of report queue is the best solution in my case, in which I set '--num_workers=28' and this may generate too much reports ；).
BTW, maybe you can unify the handle of this identical exception here [learner.py](https://github.com/alex-petrenko/sample-factory/blob/master/algorithms/appo/learner.py#L1054)

alex-petrenko:
@lanpartis fair point, I also added `safe_put` to the learner: https://github.com/alex-petrenko/sample-factory/commit/1ab38e64bd0c2ab6c8c0f1c1506a236f8cae6b28

PR #103, opened 2021-03-04
Added textured option for obstacle type sphere

tushartk:


alex-petrenko (commented):
Tushar, we don't need this change.
Textured sphere should not be a new obstacle type. It's just a visualization option. It is still a sphere from the perspective of the training system.

PR #110, opened 2021-05-13
Use a queue of free buffer indices instead of a huge static shared memory data structure

alex-petrenko:


alex-petrenko:
@erikwijmans I moved CPC into another file and only later I realized that I need to change it. Sorry about that. But the changes there are really small.

If you can take a quick look at the PR and spot any potential issues, that'd be super awesome!

PR #133, opened 2022-04-01
fix(gym): a fix for missing rendering function gym.envs.classic_control

mrzhuzhe:
add a fix for missing rendering function in 'from gym.envs.classic_control import rendering' 
due to gym new version update


alex-petrenko:
Thank you! I wasn't aware that this function is no longer in the API

PR #134, opened 2022-04-02
WIP Max entropy PPO

gautams3:


sample_factory/algorithms/appo/learner.py
```diff
@@ -397,8 +397,12 @@ def _calculate_gae(self, buffer):
         This is leftover the from previous version of the algorithm.
         Perhaps should be re-implemented in PyTorch tensors, similar to V-trace for uniformity.
         """
+        ent_coeff = self.cfg.ppo_max_entropy_coeff
```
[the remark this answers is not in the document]
krzentner: I agree with Alex's main points, max entropy is definitely also compatible (and probably useful) with V-trace.

In garage we _could_ differentiate back to the entropy through the returns, since our return calculation is differentiable, but we made doing so [an explicit error](https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/npo.py#L553).

In other words, they should not be differentiable. BTW, you might want to leave a comment pointing to [this paper](https://arxiv.org/abs/1805.00909) somewhere.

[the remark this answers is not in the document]
gautams3: @krzentner What part of the paper is relevant here? I will add that contextual info in as well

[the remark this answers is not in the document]
gautams3: @alex-petrenko yes we can add it before calling `_calculate_gae(buffer)`. But notice that I am only changing the rewards locally, not changing `buffer.rewards`. 

Do you want me to update the buffer with new rewards (it might be reserved only for task rewards, independent of entropy. I'm not sure)? If not, we can pass in `rewards+buffer.weighted_entropies` (new variable inside `buffer`) when we calculat GAE. And tell me what I should change for vtrace.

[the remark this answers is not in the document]
alex-petrenko: I think it is okay to just modify the rewards directly. Then you don't need to do anything special for v-trace - it will automatically use the new reward values.

gautams3:
@alex-petrenko I added changes to fix tensor/array reshaping bugs, in two similar ways. Please see which one works better for the coding style in SF.

sample_factory/algorithms/appo/learner.py
```diff
@@ -446,6 +445,26 @@ def _prepare_train_buffer(self, rollouts, macro_batch_size, timing):
                 if isinstance(x[0], (dict, OrderedDict)):
                     buffer[key] = list_of_dicts_to_dict_of_lists(x)
 
+        # Add max entropy to the rewards
+        with torch.no_grad():
+            ent_coeff = self.cfg.ppo_max_entropy_coeff
+            # flatten the action dist parameters in the batch
+            logits = np.array(buffer.action_logits) # [E, T, A]
```
gautams3, line 452: @alex-petrenko: changes to fix tensor/array reshaping bugs, in two similar ways. Please see which one works better for the coding style in SF.

Note that this is inefficient since we have to reshape it again to match with the shape of the rewards (`np.stack(buffer.rewards).squeeze().shape`)

alex-petrenko:
I'm looking at new changes now

sample_factory/algorithms/appo/appo_utils.py
```diff
@@ -198,21 +198,22 @@ def cat(self, dict_of_tensor_arrays, macro_batch_size, use_pinned_memory, timing
...
         else:
             with timing.add_time('batcher_mem'):
-                for d1, cache_d, key, tensor_arr, cache_t in iter_dicts_recursively(dict_of_tensor_arrays, tensor_batch):
-                    offset = 0
-                    for t in tensor_arr:
-                        first_dim = t.shape[0]
-                        cache_t[offset:offset + first_dim].copy_(torch.as_tensor(t))
-                        offset += first_dim
+                # this is slower than the older version where we copied trajectories to the cached tensor one-by-one
+                # this will only affect CPU-based envs with pixel observations, and won't matter after
```
gautams3, line 211: I'm assuming you've tested this, or have some data for which it will affect CPU-based envs only.

sample_factory/algorithms/appo/appo.py
```diff
@@ -127,42 +127,43 @@ def arg(*args, **kwargs):
...
+                 'this action ever again.'
+                 'Unlike the entropy term, symmetric KL divergence between the action distribution '
+                 'and a uniform prior approaches infinity when entropy of the distribution approaches zero,'
+                 ' so it can prevent the pathological situations where the agent stops exploring. '
+                 'Empirically, symmetric KL-divergence yielded slightly better results on some problems.',
+        )
+
+        # Max entropy objective
+        arg(
+            '--max_entropy_coeff', default=0.0, type=float,
```
gautams3, line 153: this renaming is fine, depends on what all you plan to include for SF2. This PR was only changing things for APPO

sample_factory/benchmarking/benchmarking.txt
```diff
@@ -467,6 +467,14 @@ python -m sample_factory.algorithms.appo.train_appo --env=doom_benchmark --algo=
 [2022-01-09 05:20:27,264][184143] Policy worker avg. requests 6.32, timing: init: 2.7602, wait_policy_total: 15.4553, wait_policy: 0.0052, handle_policy_step: 29.2352, one_step: 0.0025, deserialize: 3.5863, stack: 0.2062, obs_to_device: 7.6138, forward: 9.5064, to_cpu: 3.3929, format_outputs: 0.9884, postprocess: 2.5917, weight_update: 0.0006
 [2022-01-09 05:20:27,366][184116] GPU learner timing: extract: 0.1750, buffers: 0.0567, batching: 5.2542, buff_ready: 0.1724, tensors_gpu_float: 1.6330, squeeze: 0.0064, prepare: 7.1875, batcher_mem: 5.1569
 [2022-01-09 05:20:27,673][184116] Train loop timing: init: 1.9581, train_wait: 0.3825, epoch_init: 0.0011, minibatch_init: 0.0006, forward_head: 0.2330, bptt_initial: 2.9500, bptt_forward_core: 0.2780, bptt: 0.2895, tail: 0.1421, vtrace: 0.6693, losses: 0.4309, clip: 0.2765, update: 1.8503, after_optimizer: 0.3446, train: 7.3772
-[2022-01-09 05:20:27,801][184052] Workers joined!
 [2022-01-09 05:20:27,810][184052] Collected {0: 2015232}, FPS: 48606.9
-[2022-01-09 05:20:27,811][184052] Timing: experience: 41.2913
\ No newline at end of file
+[2022-01-09 05:20:27,811][184052] Timing: experience: 41.2913
+
+V1.123.0 with max entropy and GAE
```
gautams3, line 473: outside of this benchmarking, will you add release notes with this PR?

sample_factory/benchmarking/benchmarking.txt
```diff
@@ -467,6 +467,14 @@ python -m sample_factory.algorithms.appo.train_appo --env=doom_benchmark --algo=
...
+[2022-04-11 18:30:13,170][100994] Policy worker avg. requests 7.56, timing: init: 2.6132, wait_policy_total: 17.5669, wait_policy: 0.0052, handle_policy_step: 30.4024, one_step: 0.0025, deserialize: 3.7868, stack: 0.2223, obs_to_device: 8.0086, forward: 9.7811, to_cpu: 3.4902, format_outputs: 1.0286, postprocess: 2.7086, weight_update: 0.0008
+[2022-04-11 18:30:13,266][100967] GPU learner timing: extract: 0.1784, buffers: 0.0631, buffer_stack_and_squeeze: 4.7646, calc_gae: 0.4842, batching: 4.3630, buff_ready: 0.2103, tensors_gpu_float: 1.5224, prepare: 11.8780, batcher_mem: 4.2958
+[2022-04-11 18:30:13,572][100967] Train loop timing: init: 1.9517, train_wait: 0.4877, epoch_init: 0.0013, minibatch_init: 0.0006, forward_head: 0.2542, bptt_initial: 2.9018, bptt_forward_core: 0.3037, bptt: 0.3138, tail: 0.1717, losses: 0.4898, kl_divergence: 0.0629, clip: 0.2853, update: 1.9601, after_optimizer: 0.3584, train: 6.8233
+[2022-04-11 18:30:13,711][100943] Collected {0: 2015232}, FPS: 44513.3
+[2022-04-11 18:30:13,711][100943] Timing: experience: 45.0885
```
gautams3, line 480: Compared to v1.121.4, this is ~10% slower now!

gautams3 (commented):
looks good. Waiting for v1.123 now :)

sample_factory/algorithms/appo/appo_utils.py
alex-petrenko: Basically this is only slower if the learner's batcher is the bottleneck.
This only happens in rather extreme configurations (i.e. 100 CPU workers and 1 GPU learner).

Actually most of the slowdown that I measured was from slower conda environment rather than from this change.
Apparently newer versions of Numpy or MKL are quite a bit slower. The real slowdown is negligible for most envs.

sample_factory/algorithms/appo/appo.py
alex-petrenko: Old name was fine too, but the parameter itself is not PPO-specific, so shorter name should be fine.

sample_factory/benchmarking/benchmarking.txt
alex-petrenko: yes!

sample_factory/benchmarking/benchmarking.txt
alex-petrenko: Actually it is not really slower, the right comparison is vs 1.121.0 which I measured on the same system and conda env.
There there isn't any slowdown (44707 vs 44513 FPS).

PR #143, opened 2022-06-23
Sf2 mujoco

andrewzhang505:
Added train_, enjoy_ and runner for mujoco envs along with some small changes to existing code. Hyperparameter tuning for some mujoco envs still incomplete.


sample_factory/algo/sampling/batched_sampling.py
```diff
@@ -266,6 +267,9 @@ def advance_rollouts(self, policy_id: PolicyID, timing) -> Tuple[List[Dict], Lis
 
             # record the results from the env step
             rewards_cpu = rewards.cpu()
+            
+            if isinstance(infos, list):
+                infos = list_of_dicts_to_dict_of_lists(infos)
```
alex-petrenko, line 272: Actually... I see a problem with this.
The way this list_of_dicts... function is written, it will append to a list if and only if the key is present in the individual dictionary.

So if environments 1 and 3 have a key in info and environment 2 does not, we will end up with lists of length 2.
And so information for env 3 will correspond to environment 2, which will cause bugs.

Unfortunately we just can't rely on the fact that envs will return a consistent set of keys every time.
VizDoom does not. Lots of environments return a special set of keys on the last step of the episode.

Bottom line, we can't use it here.
I actually fixed the crash inside of process_rewards by this:
`time_outs = None if not isinstance(infos, dict) else infos.get('time_outs')` 
If this fixes your problem, maybe let's just use this for now?

sample_factory/algo/sampling/non_batched_sampling.py
```diff
@@ -121,7 +121,8 @@ def set_trajectory_data(self, data: Dict, rollout_step: int):
         self.curr_traj_buffer[rollout_step] = data
 
     def reset_rnn_state(self):
-        self.last_rnn_state[:] = 0.0
+        if self.cfg.use_rnn:
```
alex-petrenko, line 124: Does this serve any specific purpose, or just for clarity?
If use_rnn is False, then we initialize rnn_state tensors to be size 1 and self.last_rnn_state[:] = 0.0 is a cheap operation.
Anyways, I don't mind this change, but if this was to fix some bug please let me know or add a comment! :)

sample_factory/algo/utils/make_env.py
```diff
@@ -132,7 +132,7 @@ def step(self, actions: Tensor):
         ofs = 0
         next_ofs = self.single_env_agents
         for i, e in enumerate(self.envs):
-            idx = ofs if self.single_env_agents == 1 else slice(ofs, next_ofs)
+            idx = slice(ofs, next_ofs)
```
alex-petrenko, line 135: Can you explain this change?
I think this breaks test_example.py 
If we have a single agent we want to reduce the dimensionality, i.e. go from [[action]] to [action]
If you use slice then you will still have [[action]] which will break some envs.

Unless I am missing something here

sample_factory/envs/mujoco/mujoco_params.py
```diff
@@ -1,13 +1,80 @@
...
+        gamma=0.99,
+        gae_lambda=0.95,
+        with_vtrace=False,
+        recurrence=1,
+        value_bootstrap=True,  # assuming reward from the last step in the episode can generally be ignored
+        normalize_input=True,
+        experiment_summaries_interval=3,  # experiments are short so we should save summaries often
+        save_every_sec=15,
+
+        serial_mode=False,
```
alex-petrenko, line 41: Very cool! Did you get these parameters from baselines or rl-games? Some of them also look similar to isaacgymenvs

sample_factory/envs/mujoco/mujoco_params.py
```diff
@@ -1,13 +1,80 @@
...
+        exploration_loss_coeff=0.0,
+        learning_rate=3e-4,
+        lr_schedule='kl_adaptive_minibatch',
+        lr_schedule_kl_threshold=0.08,
+        shuffle_minibatches=True,
+        gamma=0.99,
+        gae_lambda=0.95,
+        with_vtrace=False,
+        recurrence=1,
+        value_bootstrap=True,  # assuming reward from the last step in the episode can generally be ignored
```
alex-petrenko, line 36: I think this should be False since mujoco does not return timeouts

sample_factory/envs/mujoco/mujoco_params.py
```diff
@@ -1,13 +1,80 @@
...
+        reward_scale=0.1,
+        rollout=16,
+        max_grad_norm=0.0,
+        ppo_epochs=10,
+        ppo_clip_ratio=0.2,
+        value_loss_coeff=2.0,
+        exploration_loss_coeff=0.0,
+        learning_rate=3e-4,
+        lr_schedule='kl_adaptive_minibatch',
+        lr_schedule_kl_threshold=0.08,
```
alex-petrenko, line 30: I'm fine with using adaptive LR, but stable baselines probably doesn't use this? Or do they?

sample_factory/utils/gpu_utils.py
```diff
@@ -10,7 +10,10 @@
 
 def set_global_cuda_envvars(cfg):
     if CUDA_ENVVAR not in os.environ:
-        available_gpus = get_gpus_without_triggering_pytorch_cuda_initialization(os.environ)
+        if cfg.device == 'cpu':
```
alex-petrenko, line 13: I guess that's fair! :)

sample_factory/runner/runs/mujoco_all_envs.py
```diff
@@ -0,0 +1,16 @@
...
+_experiments = [
+    Experiment(
+        'mujoco_all_envs',
+        'python -m sample_factory_examples.mujoco_examples.train_mujoco --algo=APPO',
+        _params.generate_params(randomize=False),
+    ),
+]
+
+
+RUN_DESCRIPTION = RunDescription('mujoco_all_envs', experiments=_experiments)
```
alex-petrenko, line 16: This is missing multiple seeds per run. Ideally we need to evaluate like 5-10 seeds on each env, at least for the final result.

alex-petrenko (changes requested):
Overall, awesome work! Let's make a few changes and we can merge this!
Also I'll write a few comments on Slack

sample_factory/algo/sampling/non_batched_sampling.py
andrewzhang505: When I try use_rnn = False in non-batched, I get an error of "IndexError: too many indices for array" when doing "self.last_rnn_state[:] = 0.0"

sample_factory/algo/utils/make_env.py
andrewzhang505: The MultiAgentWrapper has another reduction in dimensionallity with 'action = action[0]' which was breaking mujoco since it changed a size-N action to size 1. If we want different behaviour with a single agent, maybe we should change the multi-agent wrapper instead?

sample_factory/envs/mujoco/mujoco_params.py
andrewzhang505: They used a linearly decaying LR but I didn't see an option for that in sample factory. Is there a way to do that or should I switch to a constant LR?

andrewzhang505: RL-games uses adaptive learning rate though

sample_factory/envs/mujoco/mujoco_params.py
andrewzhang505: RL games and baselines didn't provide many parameters (mostly just LR, batch size and gamma values) so the rest I copied from isaac gym and tuned from there.

sample_factory/algo/sampling/non_batched_sampling.py
alex-petrenko: Okay, gotcha. Good find!
I believe the problem is elsewhere, we should not be fixing this line. Here: https://github.com/alex-petrenko/sample-factory/commit/4275daf2ef1cf49e68617b40dc9f2a432abbe118

I think this fixes it. I don't know why that squeeze was needed there. I tried on a couple of envs and it seems to work. Pull my changes.

sample_factory/algo/utils/make_env.py
alex-petrenko: Ok, so this is not the right way to fix it because it breaks other envs.
I'm thinking now whether we even need MultiAgentWrapper if we're using SequentialVectorize. I think I should look at this problem tomorrow and develop a proper fix.

If you can give me instructions on how to reproduce this problem, it'd be great.

I hope this does not stop you from working on evaluations

sample_factory/envs/mujoco/mujoco_params.py
```diff
@@ -1,13 +1,91 @@
...
+            learning_rate=3e-3,
+            lr_schedule_kl_threshold=0.008,
+            batch_size=1024,
+            rollout=64,
+            lr_schedule='constant',
+        )
+    if env_name == 'walker':
+        parser.set_defaults(
+            reward_scale=0.01,
+            learning_rate=0.0003,
```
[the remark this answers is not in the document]
alex-petrenko: Just double check this please.

sample_factory_examples/mujoco_examples/test_mujoco.py
```diff
@@ -0,0 +1,61 @@
...
+
+        status = run_rl(cfg)
+        self.assertEqual(status, ExperimentStatus.SUCCESS)
+
+    def test_sanity(self):
+        """
+        Run the test env in various configurations just to make sure nothing crashes or throws exceptions.
+        """
+        # Currently fails on last 2 environments
+        env_list = ["mujoco_ant", "mujoco_halfcheetah", "mujoco_hopper", "mujoco_humanoid", "mujoco_reacher", "mujoco_walker", "mujoco_swimmer", "mujoco_pendulum", "mujoco_doublependulum"]
```
alex-petrenko, line 57: Please keep 'single quotes'
We gotta set up this automatic style checker thing

alex-petrenko (approved):
Very good! Please address my comments in your next PR, but they're minor :)

PR #145, opened 2022-06-24
Sf2 auto test

wmFrank:
remove travis ci, add github actions ci and codecov support


wmFrank:
The check fails because haven't fixed all the tests yet. But the ci workflow is correct.

wmFrank:
I've verified the codecov function correctness on a small project. So I temporally commnented it until we fix all the tests.

.github/workflows/test-ci.yml
```diff
@@ -0,0 +1,38 @@
...
+        echo $CONDA/bin >> $GITHUB_PATH
+    - name: Install conda env & dependencies
+      run: |
+       conda install pytorch torchvision torchaudio cudatoolkit=11.3 -c pytorch
+       pip install -e .
+       conda list
+    - name: Run tests
+      run: |
+        ./all_tests.sh
+#    - name: Install codecov dependencies
```
alex-petrenko, line 29: Why are these commented out?

all_tests.sh
```diff
@@ -1,15 +1,15 @@
...
 
-python -m unittest sample_factory_examples.tests.test_example_multi
-STATUS_TEST_EXAMPLE_MULTI=$?
+#python -m unittest sample_factory_examples.tests.test_example_multi
+#STATUS_TEST_EXAMPLE_MULTI=$?
 
 echo "Your terminal might be unresponsive (caused by Doom envs manipulating the console). Type reset in your terminal and press Enter to get it back to normal"
-echo "Unit tests status (0 = Success): $STATUS_ALL_TESTS"
+#echo "Unit tests status (0 = Success): $STATUS_ALL_TESTS"
 echo "Custom env status (0 = Success): $STATUS_TEST_EXAMPLE"
```
alex-petrenko, line 14: Ok so we're running only one test now. This is cool! Let's fix and add more tests in the coming days! You can use `@unittest.skipUnless` and similar functions to skip individual tests.

alex-petrenko (approved):
Everything looks great! Awesome work!

Will this now run each time we commit things to sf2 branch?
I will write a few more comments on Slack

.github/workflows/test-ci.yml
wmFrank: Well, if we open it, you can refer to this link : https://github.com/wmFrank/sample-factory/runs/7034191959?check_suite_focus=true
It does show the coverage situation:
image
but it also checks all the test in the project:
image
And you can see below, it blocks the upload of the coverage report.
So we won't see any thing like this:
image

all_tests.sh
wmFrank: Yeah, that is where I am stuck, trying to fix other tests😂. Forgive me for my slowness, the code is a little hard for me. I will absolutely learn from Andrew.

PR #147, opened 2022-06-24
Fixes action distribution tests

edbeeching:
Note, I made two modifications to the TupleActionDistribution Class, please review them carefully.

Do we have an example environment with a TupleActionDistribution that I could use to further test this change?

### DIFF
--- a/sample_factory/algo/utils/action_distributions.py
+++ b/sample_factory/algo/utils/action_distributions.py
@@ -178,7 +178,7 @@ class TupleActionDistribution:
 
     @staticmethod
     def _flatten_actions(list_of_action_batches):
-        batch_of_action_tuples = torch.stack(list_of_action_batches).transpose(0, 1)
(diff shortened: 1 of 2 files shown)


sample_factory/algo/utils/action_distributions.py
```diff
@@ -206,7 +206,6 @@ def sample(self):
     def log_prob(self, actions):
         # split into batches of actions from individual distributions
         list_of_action_batches = torch.chunk(actions, len(self.distributions), dim=1)
-        list_of_action_batches = [a.squeeze(dim=1) for a in list_of_action_batches]
```
alex-petrenko, line 209: We'll have to make sure that this does not break any of the environments.
VizDoom has multi-head action spaces in envs like Battle, this is where we can test it.

alex-petrenko (approved):

PR #150, opened 2022-06-27
[SF2] Adds support and tests for mixed Action Distributions

edbeeching:
We should perform more detailed testing by training on compatible or dummy env.

### DIFF
--- a/sample_factory/algo/utils/action_distributions.py
+++ b/sample_factory/algo/utils/action_distributions.py
@@ -14,7 +14,7 @@ def calc_num_actions(action_space):
     if isinstance(action_space, gym.spaces.Discrete):
         return 1
     elif isinstance(action_space, gym.spaces.Tuple):
-        return len(tuple(action_space.spaces))
+        return sum([calc_num_actions(a) for a in action_space])
     elif isinstance(action_space, gym.spaces.Box):
         if len(action_space.shape) != 1:
             raise Exception('Non-trivial shape Box action spaces not currently supported. Try to flatten the space.')
@@ -29,7 +29,7 @@ def calc_num_logits(action_space):
     if isinstance(action_space, gym.spaces.Discrete):
         return action_space.n
     elif isinstance(action_space, gym.spaces.Tuple):
-        return sum(space.n for space in action_space.spaces)
+        return sum([calc_num_logits(a) for a in action_space])
     elif isinstance(action_space, gym.spaces.Box):
         # regress one mean and one standard deviation for every action
         return np.prod(action_space.shape) * 2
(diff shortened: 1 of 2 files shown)


sample_factory/algo/utils/action_distributions.py
```diff
@@ -14,7 +14,7 @@ def calc_num_actions(action_space):
     if isinstance(action_space, gym.spaces.Discrete):
         return 1
     elif isinstance(action_space, gym.spaces.Tuple):
-        return len(tuple(action_space.spaces))
+        return sum([calc_num_actions(a) for a in action_space])
```
alex-petrenko, line 17: Nice! :)

tests/test_action_distribution.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+import gym
+import numpy as np
+import torch
+from sample_factory.algo.utils.action_distributions import calc_num_actions, calc_num_logits, get_action_distribution
+
+@pytest.mark.parametrize("spaces", [
```
alex-petrenko, line 7: Single-quote policy throughout the codebase! :)

tests/test_action_distribution.py
```diff
@@ -0,0 +1,54 @@
...
+            [gym.spaces.Discrete, gym.spaces.Discrete],
+            [gym.spaces.Discrete, gym.spaces.Box],
+            [gym.spaces.Box, gym.spaces.Box]
+        ])
```
alex-petrenko, line 11: I really like this. Pytest is great

tests/test_action_distribution.py
```diff
@@ -0,0 +1,54 @@
...
+    action_log_probs = action_dist.log_prob(actions)
+    assert action_log_probs.size() == (BATCH_SIZE, )
+
+    entropy = action_dist.entropy()
+    assert entropy.size() == (BATCH_SIZE, )
+
+    actions, action_log_probs = action_dist.sample_actions_log_probs()
+
+    assert actions.size() == (BATCH_SIZE, num_actions)
+    assert action_log_probs.size() == (BATCH_SIZE, )
```
alex-petrenko, line 54: Since new tests use pytest (and I clearly see a lot of value in using it), let's prioritize switching the rest of the tests to PyTest. Because as far as I understand, this test is not going to run with the current CI system.

@wmFrank @edbeeching I think we should do it before we start working on the documentation website. Can you prioritize this?

alex-petrenko (commented):
Very cool!

I guess we can train a VizDoom env with a Discrete + Continuous distribution, I think Battle is a good candidate.

@edbeeching you mentioned that your have you own environment that uses such action spaces. Do you think we should add it as an example?

alex-petrenko (approved):

PR #155, opened 2022-07-01
Sf2 normalize returns

alex-petrenko:
Finished working on return normalization.

This PR contains a few other changes as well, most important one is switching to gae returns (returns = adv + values)


alex-petrenko:
image
Results on ige Ant
With normalized returns sample efficiency is a bit worse, but this is expected. Return normalization effectively makes the critic loss a lot smaller (since value targets are small numbers). This is similar to having a much smaller critic loss coefficient.
Also on such short training sessions the critic will need to adapt to changing statistics (i.e. the same env state will result in different value estimate at different stages in training because of RMS normalization - and that's okay)

The results in rl_games are exactly the same - training with return normalization on Ant is a bit slower. But with normalization we don't need to tune reward scale

alex-petrenko:
It even improves results on Humanoid actually 
image

alex-petrenko:
Also value loss graph looks better at convergence

image

wmFrank (approved):

PR #156, opened 2022-07-01
Fixes some of the issues with running the DMlab30 benchmark

edbeeching:
Outstanding items to resolve:

1. handlers/callbacks for EXTRA_EPISODIC_STATS_PROCESSING & EXTRA_PER_POLICY_SUMMARIES
2. obtain_env_info_in_a_separate_process() creates a cache of the env info on disk, if we use this cache the sf_context object is not updated with the custom model definition, as this is registered when a dummy env is created. I commented out this cache for the moment.
3. PBT - I switched off all the arguments related to this as it is not implemented yet.

### DIFF
--- a/sample_factory/algo/learning/learner.py
+++ b/sample_factory/algo/learning/learner.py
@@ -804,8 +804,12 @@ class Learner(EventLoopObject, Configurable):
             # we still reference the same buffers though
             buff = copy.copy(self.batcher.training_batches[batch_idx])
 
-            # TODO: how about device_and_type_for_input_tensor
-
+            # Some inputs such as DMLab text embeddings are faster to calculate on the CPU
+            for k, v in buff['obs'].items():
+                device, dtype = self.actor_critic.device_and_type_for_input_tensor(k)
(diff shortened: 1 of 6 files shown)


alex-petrenko:
Summary of what I tried to convey in PR comments in the code (read them first :) )

> handlers/callbacks for EXTRA_EPISODIC_STATS_PROCESSING & EXTRA_PER_POLICY_SUMMARIES
This should be done through new custom API on the runner

> obtain_env_info_in_a_separate_process() creates a cache of the env info on disk, if we use this cache the sf_context object is not updated with the custom model definition, as this is registered when a dummy env is created. I commented out this cache for the moment.
Let's keep and improve env info cache. DMLab issue is unrelated and should be solved by directly modifying sf_context in train_dmlab.py

> PBT - I switched off all the arguments related to this as it is not implemented yet.
That's fine. These are remnants from sf1. Refactoring was a bit too hard in places if I were to keep all of the features. I think bringing back PBT will be relatively easy, we can uncomment the cfg later.

edbeeching:
Superceded by #166

PR #159, opened 2022-07-04
Implements Tuple action dist training in non-batched mode

edbeeching:
Fixes the output actions for non-batched mode.

I am not sure how to pass command line args to a pytest test, I could add these to the bash script, but it might be nice to parameterize these in future tests.

Example script to check it works:
python tests/test_mixed_dist_env.py --env mix_dist_env --experiment mix_dist_env

### DIFF
--- a/sample_factory/algo/sampling/non_batched_sampling.py
+++ b/sample_factory/algo/sampling/non_batched_sampling.py
@@ -2,9 +2,11 @@ import random
 from queue import Empty
 from typing import Tuple, Dict, List, Optional
 
+import gym
 import numpy as np
 
 from sample_factory.algo.sampling.sampling_utils import VectorEnvRunner, TIMEOUT_KEYS, fix_action_shape
+from sample_factory.algo.utils.action_distributions import calc_num_actions
 from sample_factory.algo.utils.env_info import EnvInfo
 from sample_factory.algo.utils.make_env import make_env_func_non_batched
 from sample_factory.algo.utils.tensor_dict import to_numpy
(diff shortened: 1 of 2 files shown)


tests/test_mixed_dist_env.py
```diff
@@ -0,0 +1,99 @@
+from typing import Optional, Tuple, Union, List
```
alex-petrenko, line 1: We should have it either as a test or a proper example :)

sample_factory/algo/sampling/non_batched_sampling.py
```diff
@@ -125,11 +127,36 @@ def curr_actions(self) -> np.ndarray:
         :return: the latest set of actions for this actor, calculated by the policy worker for the last observation
         """
         actions = ensure_numpy_array(self.last_actions)
-        if self.env_info.integer_actions:
+
+        if isinstance(self.env_info.action_space, gym.spaces.Discrete):
```
alex-petrenko, line 131: Looks like this can be just `isinstance(action_space, [Discrete, Box])` instead of two separate ifs

alex-petrenko: or since we're checking for the type here, maybe we can then avoid a type check inside calculate_actions()

sample_factory/algo/sampling/non_batched_sampling.py
```diff
@@ -125,11 +127,36 @@ def curr_actions(self) -> np.ndarray:
...
+            return self.calculate_actions()
+
+        if isinstance(self.env_info.action_space, gym.spaces.Box):
+            return self.calculate_actions()
+
+
+        if isinstance(self.env_info.action_space, gym.spaces.Tuple):
+            # calculate splits
+            splits = [calc_num_actions(space) for space in self.env_info.action_space]
```
alex-petrenko, line 140: I think this solution is good, except maybe I'd cache things like calc_num_actions in the ctor to avoid doing this every time.

Also, may I suggest an alternative solution?
What about if we have mixed actions we just leave them all to be float and the env can then convert some of the actions to ints that need to be converted?
The rationale is that the env knows which actions are which type anyway.

Since you're the author of the env with mixed actions, you're a better authority on this, I don't feel strongly either way :)

alex-petrenko (commented):
I think overall I'm fine with this approach! 
Do you want me to merge this or add the batched sampling version first?

edbeeching:
Replaced by #165

PR #162, opened 2022-07-07
Sf2 mujoco

andrewzhang505:
Updated mujoco params and README.
Added linear decay learning rate scheduler.
Moved files from sample_factory.envs.mujoco to sample_factory_examples.mujoco_examples
Updated mujoco tests - currently failing single-action-envs with batched sampling - looking into it


sample_factory/algo/learning/learner.py
```diff
@@ -111,13 +111,30 @@ def invoke_after_each_epoch(self):
         return True
 
 
+class LinearDecayScheduler(LearningRateScheduler):
+    def __init__(self, cfg):
+        num_updates = cfg.train_for_env_steps // cfg.batch_size * cfg.num_epochs
+        self.linear_decay = LinearDecay([(0, cfg.learning_rate), (num_updates, 0)])
```
alex-petrenko, line 117: Should we decay to 0 or to some really small value like 1e-6?
What do the other frameworks do?

alex-petrenko (approved):
The report looks awesome!
I think this is very good for now and we should move to other envs.

Remaining things for Mujoco that we can do later:
1) Do another report where we compare wall-time performance on Mujoco vs SB3 or CleanRL. We should try to beat both envpool and non-envpool versions by a significant margin - and this is actually what we're going to show people :) This should be motivation for ppl to use SF2 over other frameworks.

2) Upload trained models to HuggingFace (I know @edbeeching is helping you with this!)

3) Setup some kind of CI that repeatedly trains on mujoco, possibly on HF compute? :)

PR #163, opened 2022-07-08
fix enjoy vizdoom error

wmFrank:


alex-petrenko:
Should we fix the tests first before I merge it?

wmFrank:
> Should we fix the tests first before I merge it?

Absolutely I should fix it before merging. We must use the newest gym version, otherwise mujoco will not work. Instead of rendering, newest gym uses pygame, so I am working on changing the vizdoom render function to pygame.

wmFrank:
Yeah, I think it is ready to be merged.

sample_factory/envs/doom/doom_gym.py
```diff
@@ -441,17 +442,26 @@ def render(self, mode="human"):
...
+            import pygame
+
+            if self.screen is None:
+                pygame.init()
+                pygame.display.init()
+                self.screen = pygame.display.set_mode((render_w, render_h))
+
+            self.surf = pygame.surfarray.make_surface(np.rot90(img))
+            self.screen.blit(self.surf, (0, 0))
+            pygame.display.update()
```
alex-petrenko, line 462: Ok, good solution!

sample_factory/envs/doom/doom_gym.py
```diff
@@ -463,8 +473,13 @@ def close(self):
         except RuntimeError as exc:
             log.warning("Runtime error in VizDoom game close(): %r", exc)
 
-        if self.viewer is not None:
-            self.viewer.close()
+        # if self.viewer is not None:
+        #     self.viewer.close()
```
alex-petrenko, line 477: Unless it serves a purpose, try not to leave commented code like this. To find old code we can always use git :)

setup.py
```diff
@@ -47,8 +47,8 @@
     extras_require={
         'dev': ['black', 'isort', 'pytest<8.0', 'flake8'],
         'atari': ['gym[atari, accept-rom-license]'],
-        'mujoco': ['gym[mujoco]<1.0', 'mujoco', 'mujoco_py>=2.0,<3.0'],
-        'vizdoom': ['vizdoom<2.0'],
+        'mujoco': ['gym[mujoco, mujoco_py]'],
+        'vizdoom': ['vizdoom<2.0', 'gym[classic_control]'],
```
alex-petrenko, line 51: I'm assuming this is for pygame dependency? Cool.

alex-petrenko (approved):

setup.py
wmFrank: Yes, it is for pygame

PR #165, opened 2022-07-09
Implements Tuple action distributions for non-batched and batched training

edbeeching:
Improves on #159 

Retains old behavior when the Tuple action space contains only discrete spaces.

Tests:

pytest tests/test_mixed_dist_env_batched.py
pytest tests/test_mixed_dist_env_non_batched.py
pytest tests/test_two_discrete_action_dist_env_batched.py
pytest tests/test_two_discrete_action_dist_env_non_batched.py

### DIFF
--- a/.github/workflows/cover-ci.yml
+++ b/.github/workflows/cover-ci.yml
@@ -22,8 +22,6 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install conda env & dependencies
       run: |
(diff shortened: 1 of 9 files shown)


edbeeching:
> I was a bit confused about fix_action_shape since I thought I removed this function, but then I realized that the function is still there, even though code in sf2 does not use it. So I removed all code that used this function, but forgot to delete the function itself.

I think the function must have been removed after I started on this branch. Perhaps the code can be simplified further?

alex-petrenko (approved):

PR #166, opened 2022-07-11
DM-lab training example

edbeeching:
Moves to the examples folder

Creates a train.py example for the DMLab30 benchmark.

Still TODO:
1. Enjoy.py file raises an error based on obs not being a dict. Perhaps the dict auto wrapper is not working?
2. Summary handlers for EXTRA_EPISODIC_STATS_PROCESSING, EXTRA_PER_POLICY_SUMMARIES


sample_factory_examples/dmlab/dmlab_env.py
```diff
@@ -3,21 +3,24 @@
...
-from sample_factory.envs.dmlab.wrappers.reward_shaping import RAW_SCORE_SUMMARY_KEY_SUFFIX, DmlabRewardShapingWrapper
-from sample_factory.envs.env_wrappers import PixelFormatChwWrapper, RecordingWrapper
-from sample_factory.utils.utils import experiment_dir, log, static_vars
+from sample_factory_examples.dmlab.dmlab_gym import DmlabGymEnv, dmlab_level_to_level_name
+from sample_factory_examples.dmlab.dmlab_level_cache import dmlab_ensure_global_cache_initialized
+from sample_factory_examples.dmlab.dmlab_model import dmlab_register_models
+from sample_factory_examples.dmlab.dmlab_utils import EXTRA_EPISODIC_STATS_PROCESSING, EXTRA_PER_POLICY_SUMMARIES
+from sample_factory_examples.dmlab.wrappers.reward_shaping import (
+    RAW_SCORE_SUMMARY_KEY_SUFFIX,
+    DmlabRewardShapingWrapper,
```
alex-petrenko, line 22: These imports need to be updated.

sample_factory_examples/dmlab/dmlab_env.py
```diff
@@ -147,11 +150,11 @@ def make_dmlab_env_impl(spec, cfg, env_config, **kwargs):
     return env
 
 
-def make_dmlab_env(env_name, cfg=None, **kwargs):
+def make_dmlab_env(env_name, cfg, env_config, **kwargs):  # TODO: env_config required here for DMlab?
```
alex-petrenko, line 153: make_env_func should take 3 arguments.
See the new function `register_env()` and the docstring comments. Also new type CreateEnvFunc used for typehints

sample_factory_examples/dmlab/dmlab_level_cache.py
```diff
@@ -218,9 +218,9 @@ def add_new_level(self, level, seed, key, pk3_path):
 def dmlab_ensure_global_cache_initialized(experiment_dir, all_levels_for_experiment, num_policies, level_cache_dir):
     global DMLAB_GLOBAL_LEVEL_CACHE
 
-    assert (
-        multiprocessing.current_process().name == "MainProcess"
-    ), "make sure you initialize DMLab cache before child processes are forked"
+    # assert (
+    #     multiprocessing.current_process().name == "MainProcess"
+    # ), "make sure you initialize DMLab cache before child processes are forked"
```
alex-petrenko, line 223: There should be a way to propagate cache across processes, i.e. via sf_global_context()
So I believe this assert is still relevant?

sample_factory_examples/dmlab/dmlab_params.py
```diff
@@ -4,20 +4,26 @@
...
+        hidden_size=256,
         obs_subtract_mean=0.0,
+        nonlinearity="relu",
         obs_scale=255.0,
         env_frameskip=4,
+        rollout=32,
+        recurrence=32,
+        rnn_type="lstm",
+        max_policy_lag=35,
+        num_epochs=1,
```
alex-petrenko, line 21: Are these changes from the original IMPALA paper? Why do we decrease the hidden size?

edbeeching: This is a comparable to the arguments you provide in the README. Should the example match the README, or would you like something else?

sample_factory_examples/dmlab/dmlab_level_cache.py
edbeeching: Whenever I ran the example this assert would be triggered. I will take a look into it.

alex-petrenko: The thing is in SF2 we can't just store data in global variables. If we want to use global state it should be communicated through sf_global_context() because this context is passed on explicitly when we create new processes.
Or we should use a different mechanism that does not rely on global state, not sure what is better.

Cache mechanism currently uses a bunch of global state I believe and therefore needs to be reworked. Probably initializing it explicitly and then passing along through global context is the easiest way to do this :)

This all worked in SF1 because it used "fork" multiprocessing method, but this created lots of other limitations, such as not being able to use CUDA in the main process and problems with envs like isaacgym.

alex-petrenko:
Merged most of the changes manually directly in sf2 branch. Also added custom summary handlers for DMLab-30

PR #170, opened 2022-07-14
Sf2 slurm

andrewzhang505:
Added instructions for running slurm to docs
Fixed installation for mujoco

### DIFF
--- a/docs/get-started/running-slurm.md
+++ b/docs/get-started/running-slurm.md
@@ -1,5 +1,7 @@
 ## How to use SF2 on Slurm
 
+This doc contains instructions for running Sample-Factory v2 on USC's brain cluster
+
 ### Logging in
 
 Login to `brain.usc.edu` using ssh with your username and password.
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
     extras_require={
         "dev": ["black", "isort", "pytest<8.0", "flake8", "pre-commit"],
         "atari": ["gym[atari, accept-rom-license]"],
-        "mujoco": ["gym[mujoco, mujoco_py]"],
+        "mujoco": ["gym[mujoco]", "mujoco", "mujoco_py>=2.0,<3.0"],
         "vizdoom": ["vizdoom<2.0", "gym[classic_control]"],
     },
     package_dir={"": "./"},
--- a/docs/get-started/slurm_setup.md
+++ b/docs/get-started/slurm_setup.md
@@ -1,0 +1,42 @@
+## How to use SF2 on Slurm
+
+### Logging in
+
+Login to `brain.usc.edu` using ssh with your username and password.
+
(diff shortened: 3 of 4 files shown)


setup.py
```diff
@@ -45,7 +45,7 @@
     extras_require={
         "dev": ["black", "isort", "pytest<8.0", "flake8", "pre-commit"],
         "atari": ["gym[atari, accept-rom-license]"],
-        "mujoco": ["gym[mujoco, mujoco_py]"],
+        "mujoco": ["gym[mujoco]", "mujoco", "mujoco_py>=2.0,<3.0"],
```
wmFrank, line 48: Why do we need to set the version? I think gym will automatically import the latest mujoco.

wmFrank (approved):

setup.py
andrewzhang505: The 2nd mujoco and mujoco_py are not from gym and need to be installed seperately. The versioning on mujoco_py is probably not necessary but I included it anyways

alex-petrenko: I wonder why tests worked without it :)

Btw I think restricting the major version of a package is a good practice @wmFrank 
The reason is that major version updates usually introduce incompatible API changes. Unless we want our software to break every time authors of another package release a new major version, we should use limits like this.

One example is going to be sample factory 2.0. If someone has just sample_factory in their dependencies, us releasing a new PyPI package is going to break their library.

docs/get-started/running-slurm.md
```diff
@@ -0,0 +1,44 @@
...
+Activate your conda environment with `bash` and `conda activate sf2` then `cd sample-factory`
+
+Run your runner script - an example mujuco runner (replace run and slurm_workdir with appropriate values)
+```
+python -m sample_factory.runner.run --runner=slurm --slurm_workdir=./slurm_mujoco --experiment_suffix=slurm --slurm_gpus_per_job=1 --slurm_cpus_per_gpu=16 --slurm_sbatch_template=./sample_factory/runner/slurm/sbatch_template.sh --pause_between=1 --slurm_print_only=False --run=sf_examples.mujoco_examples.experiments.mujoco_all_envs
+```
+
+### Other Helpful Commands
+
+`squeue` - List all jobs running or queued in the cluster
```
alex-petrenko, line 44: @andrewzhang505 sorry that I wasn't completely clear about this.
I think we should have an instruction for running SF2 on Slurm clusters, not specifically on brain :)
Can you please update this instruction to not mention brain or USC, but be generally for Slurm?

We can keep this instruction for internal use, or you can share it with our labmates on compute-forum!

I will merge your PR anyway for the Mujoco change but please update it in another PR!
Also, while we're at it, can you please add documentation website link to readme? I want to see how this looks on the website :)

alex-petrenko (approved):

PR #172, opened 2022-07-15
Fixed sample counting bug

andrewzhang505:
Fixed sample counting bug caused by double counting in non-batched mode in line 198 and line 277.

Reverted mujoco installation change as it broke test. But, I cannot replicate the old installation working on my machine. Will look into it more tomorrow.

### DIFF
--- a/sample_factory/algo/inference/inference_worker.py
+++ b/sample_factory/algo/inference/inference_worker.py
@@ -195,7 +195,6 @@ class InferenceWorker(EventLoopObject, Configurable):
                 for env_idx, agent_idx, traj_buffer_idx, rollout_step in request_data:
                     index = [traj_buffer_idx, rollout_step]
                     indices.append(index)
-                    self.total_num_samples += 1
 
             indices = tuple(np.array(indices).T)
             traj_tensors = self.traj_tensors[device]  # TODO: multiple sampling devices?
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -45,7 +45,7 @@
     extras_require={
         "dev": ["black", "isort", "pytest<8.0", "flake8", "pre-commit"],
         "atari": ["gym[atari, accept-rom-license]"],
-        "mujoco": ["gym[mujoco]", "mujoco", "mujoco_py>=2.0,<3.0"],
+        "mujoco": ["gym[mujoco]"],
```
andrewzhang505, line 48: Removing mujoco_py fixes the error -> only used for deprecated mujoco versions no longer supported by gym

sample_factory/algo/inference/inference_worker.py
```diff
@@ -195,7 +195,6 @@ def _batch_individual_steps(self, timing):
                 for env_idx, agent_idx, traj_buffer_idx, rollout_step in request_data:
                     index = [traj_buffer_idx, rollout_step]
                     indices.append(index)
-                    self.total_num_samples += 1
```
alex-petrenko, line 198: Yep, that's my mistake.
Thank you for fixing!

alex-petrenko (approved):

Issue #160, opened 2022-07-04
Add make format and make check-codestyle to git pre-commit


alex-petrenko:
I recently updated the whole codebase to use black and isort formatting @edbeeching 

Since we've now automated code formatting, it makes sense to do it every time code is being committed to VCS.
Here's a short instruction I found that may be helpful: https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/

If we could make it such that anyone who clones the repo automatically gets the pre-commit hook, it'd be awesome.
@wmFrank since you are a king of CI, can you please do this thing? Not a very high priority, but ideally this should be done in the next few days.

wmFrank:
Sounds good, I'll try it.

alex-petrenko:
We also need to add `make check-codestyle` to CI checks alongside tests :)

PR #164, opened 2022-07-09
Sf2 pre-commit

wmFrank:
- add pre-commit config to automatically check and correct format and code style before every commit
- add pre-commit CI config to check format and code style after every commit
- modify CONTRIBUTING.md, add instructions to enable pre-commit
- add pre-commit badges

PR #173, opened 2022-07-20
Added pre-commit hook for black and isort

usamar240:
I have added a pre-commit hook using the pre-commit package. The file ".pre-commit-config.yaml" contains the configuration for the hook and now every time someone commits it would lint the program using black and isort. I also ran it once on all files so that the code is fully black and isort compliant. 

I have also updated the environment.yml, to include pre-commit as a dependency. You might need to run `pre-commit install` the first time.


usamar240:
Never mind I just saw the sf2 branch, which has the this already set up. I will leave it open in case you also want to include it in master branch. 

Anyway, I suggest please close issue #160 because it has already been implemented in sf2 branch in pull request #164.

PR #177, opened 2022-07-23
Sf2 atari benchmark

wmFrank:
finished atari benchmark, report here: https://wandb.ai/wmfrank/atari-benchmark/reports/Atari-Sample-Factory2-Baselines-CleanRL--VmlldzoyMzEyNjIw
- change atari wrappers
- add cfg to control whether to multiply frameskip for train steps
- add cfg to control whether to use gym RecordEpisodeStatistics to keep track of reward
- add atari benchmark configs


sample_factory/algo/sampling/non_batched_sampling.py
```diff
@@ -494,14 +503,20 @@ def _process_policy_outputs(self, policy_id, timing):
...
-    def _process_rewards(self, rewards, env_i):
+    def _process_rewards(self, rewards, infos, env_i):
         """
         Pretty self-explanatory, here we record the episode reward and apply the optional clipping and
         scaling of rewards.
         """
         for agent_i, r in enumerate(rewards):
             self.actor_states[env_i][agent_i].last_episode_reward += r
 
+        if self.cfg.use_record_episode_statistics:
```
alex-petrenko, line 514: Can we do all of this in one function "record_env_step" ?
Also, it looks like there is no point checking this on every step, can't we just look at info["episode"] when done=True like it's done in record_env_step() for "extra stats"?

sample_factory/algo/sampling/non_batched_sampling.py
```diff
@@ -86,6 +86,9 @@ def __init__(
         self.last_episode_reward = 0
         self.last_episode_duration = 0
 
+        self.episode_return = 0
```
alex-petrenko, line 89: I am not a fan of having to introduce another two variables that semantically do exactly the same as last_episode_reward and last_episode_duration. Can we just keep and reuse the old variables?

sample_factory/algo/sampling/non_batched_sampling.py
alex-petrenko: Also, we might need to implement the same thing in batched mode.
Look at _process_env_step there

sample_factory/cfg/cfg.py
```diff
@@ -544,6 +544,15 @@ def add_default_env_args(p: ArgumentParser):
     p.add_argument(
         "--pixel_format", default="CHW", type=str, help="PyTorch expects CHW by default, Ray & TensorFlow expect HWC"
     )
+    p.add_argument(
+        "--multiply_frameskip", default=True, type=str2bool, help="Whether to multiply frameskip for training steps"
```
alex-petrenko, line 548: Let's rename this to "summaries_use_frameskip" and put it into "# reporting summaries" section in cfg.py

Help string should be: "Whether to multiply training steps by frameskip when recording summaries, FPS, etc. When this flag is set to True, x-axis for all summaries corresponds to the total number of simulated steps, i.e. with frameskip=4 the x-axis value of 4 million will correspond to 1 million frames observed by the policy."

sf_examples/atari_examples/atari/atari_utils.py
```diff
@@ -38,37 +41,22 @@ def atari_env_by_name(name):
...
+    env = NoopResetEnv(env, noop_max=30)
+    env = MaxAndSkipEnv(env, skip=cfg.env_frameskip)
+    env = EpisodicLifeEnv(env)
+    if "FIRE" in env.unwrapped.get_action_meanings():
+        env = FireResetEnv(env)
+    env = ClipRewardEnv(env)
+    env = gym.wrappers.ResizeObservation(env, (84, 84))
+    env = gym.wrappers.GrayScaleObservation(env)
+    env = FrameStack(env, cfg.env_framestack)
+    # env = gym.wrappers.FrameStack(env, 4)
```
alex-petrenko, line 61: @wmFrank can you please check if any of the wrappers in envs/env_wrappers.py are currently not being used? In that case, just delete them :)

alex-petrenko (approved):

alex-petrenko (changes requested):
Very nice PR!
I requested a few small changes, please take a look!

sample_factory/algo/sampling/non_batched_sampling.py
wmFrank: Will take a look at that.

sample_factory/cfg/cfg.py
wmFrank: Ok, I'll do that

sf_examples/atari_examples/atari/atari_utils.py
wmFrank: ok.

wmFrank:
I tried putting it in "record_env_step" when "done == True", didn't work, so I changed it back. As for the "introducing another two variables" issue, I tried merging them, but it didn't work, I think the reason is that they are calculated differently and have an effect on training. But for cleanness, I add an if statement to the initialization of the variables. Also added support for batched_sampling mode.

wmFrank:
The comparisons are on WandB.

alex-petrenko:
> I tried putting it in "record_env_step" when "done == True", didn't work, so I changed it back. As for the "introducing another two variables" issue, I tried merging them, but it didn't work

Sorry, "it didn't work" is not a very good argument here. We need two variables to keep track of reward and episode length, not four. Also typically having different variables named "episode_rew" and "episode_reward" makes code very confusing. Since it's already difficult to read, I think we should make an effort to minimize any extra complexity as much as possible.

Anyway, let me know if you are finished with the PR! I can merge it and take a look at these small issues myself. Overall, great work reproducing the Atari results! :) Let's finish this and move to new tasks! :)

wmFrank:
It is ready to be merged. By "didn't work", I mean we couldn't get the expected reward.

wmFrank:
I just found some bugs about the framestack wrapper. Please do not merge it now.

alex-petrenko:
Thank  you for letting me know! What kinds of bugs?

wmFrank:
The observation shape isn't what we expect. I modified the gym stack wrapper to solve the lazy frame problems, but I may not solve them correctly. I will take a look at it.

alex-petrenko:
Sorry for asking again, but why can't we use the same wrappers as stable-baselines / clean-rl?

wmFrank:
There are two ways. 1. modify gym wrapper to convert LazyFrames type to NumPy / torch.Tensor type. 2. do not modify wrapper, convert LazyFrames to NumPy / torch.Tensor type in TensorDict. I used the second way now, so that we can use exactly the same wrappers as stable-baselines / clean-rl.

sf_examples/atari_examples/atari/atari_params.py
```diff
@@ -0,0 +1,52 @@
...
+        encoder_extra_fc_layers=1,
+        obs_subtract_mean=0.0,
+        obs_scale=255.0,
+        gamma=0.99,
+        env_frameskip=4,
+        env_framestack=4,
+        exploration_loss_coeff=0.01,
+        num_workers=8,
+        num_envs_per_worker=1,
+        worker_num_splits=1,
```
alex-petrenko, line 18: We should also do another pass at some point and provide default parameters for Atari that yield high throughput.

alex-petrenko (approved):
Great work!
I will merge it into sf2 now and do a pass with some small changes.

PR #183, opened 2022-07-29
Sf2 model builder

edbeeching:
Initial implementation of an automatic model builder.
Adds a default mode, which will generate encoder(s) automatically based on the shape(s) of the input observation(s).
Adds a decoder option after the core of the model.

I have included limited testing, but this needs to be extended.

There are some major limitations that need to be addressed, for example image observations that are too small. e.g. 32x32 require a smaller CNN, which would need to be automatically selected.


alex-petrenko:
Added this feature manually. Thank you for starting this work!

PR #189, opened 2022-08-08
Sf2 heartbeat

andrewzhang505:
Added heartbeat mechanism in the runner that receives signals from other components periodically. If a component does not send a heartbeat after `cfg.heartbeat_reporting_interval`, logs an error and if all components of a same type fail then it stops the run.

Also updated some minimum requirements in setup.py. torch v1.9 required because we use `torch.inference_mode()` and python v3.8 required because fstring self-documenting expressions.


sample_factory/algo/runners/runner.py
```diff
@@ -490,11 +495,48 @@ def _on_start(self):
...
+        self.heartbeat_dict[component_type][component_id] = curr_time
 
-    def _setup_component_termination(self, stop_signal: signal, component_to_stop: StoppableEventLoopObject):
+    def _check_heartbeat(self):
+        """
+        Reports components whose last heartbeat signal is longer than self.heartbeat_report_sec.
+        If all components of the same time fail, stop the run
+        """
+        curr_time = time.time()
+        log.info("Checking heartbeat")
```
alex-petrenko, line 526: I think this log message is redundant, let's remove it to avoid spam

sample_factory/algo/runners/runner.py
```diff
@@ -490,11 +495,48 @@ def _on_start(self):
...
+        for component_type, heartbeat_dict in self.heartbeat_dict.items():
+            num_components = len(heartbeat_dict)
+            num_stopped = 0
+            for component_id, heartbeat_time in heartbeat_dict.items():
+                if curr_time - heartbeat_time > self.heartbeat_report_sec:
+                    log.error(f"No heartbeat for {curr_time - heartbeat_time} sec from {component_id}")
```
alex-petrenko, line 532: Can you make it such that there's a single message reporting all failed components, and only print it if there's at least one.

Like: No heartbeat for components: RolloutWorker-0 (300 seconds), ... etc.

sample_factory/algo/runners/runner.py
```diff
@@ -535,8 +577,13 @@ def connect_components(self):
             self._setup_component_termination(self.stop, batcher)
             self._setup_component_termination(batcher.stop, learner)
 
+            # Heartbeats
+            self._setup_component_heartbeat(batcher)
+            self._setup_component_heartbeat(learner)
+
         for sampler_component in sampler.stoppable_components():
```
alex-petrenko, line 584: Function name "stoppable components" is a bit confusing. Maybe let's have two functions "stoppable components" and "heartbeat components"? And internally just return the same list?

Also, abstract sampler itself should probably have heartbeat too. Current sampler object itself runs on main thread but the interface allows samplers that run in a separate process (i,e, you can imagine a synchronous sampler that does inference and rollout on the same process). 

Although this is not very important for now since we don't have any other implementations.

alex-petrenko (approved):
Overall, great work as always! I will merge since this is already super useful.
Take a look at my small remarks, maybe let's fix in next PR!

PR #190, opened 2022-08-09
Sf2 cov

wmFrank:
add a config file to define the code coverage range


.core-coveragerc
```diff
@@ -0,0 +1,10 @@
+[run]
+omit =
+    setup.py
+    tests/*
```
alex-petrenko, line 4: Aren't tests covered by tests themselves? What's the purpose of removing tests?

.core-coveragerc
```diff
@@ -0,0 +1,10 @@
...
+    sample_factory/runner/runs/*
+    sample_factory/huggingface/*
```
alex-petrenko, line 6: I think we can keep huggingface and PBT. We should only remove folders that we can't test in CI such as doom or isaacgym. But PbT and huggingface we can write tests for. So coverage below 100% should reflect that! :)

Makefile
```diff
@@ -56,3 +56,10 @@ test:
 test-cov:
 	pytest --cov=./ -v
 # ; echo "Tests finished. You might need to type 'reset' and press Enter to fix the terminal window"
+
+# Run code coverage test
```
alex-petrenko, line 60: :+1:

alex-petrenko (changes requested):
Please see my minor comments!

.core-coveragerc
wmFrank: We write tests for sample-factory code, but we do not write tests for our tests.

wmFrank: Correct me if I am wrong.

.core-coveragerc
wmFrank: OK, got it.

alex-petrenko (approved):

PR #191, opened 2022-08-09
fix log missing problems

wmFrank:
It turned out that before we could only log info in the main process(so errors in other processes will not be logged). I changed the time of initializing the file logger and it works. The problem is that we cannot pass in the cfg.experiment as before. Now the sf_log.txt will be in the same dir as the train script.

### DIFF
--- a/sample_factory/algo/learning/batcher.py
+++ b/sample_factory/algo/learning/batcher.py
@@ -11,7 +11,7 @@ from sample_factory.algo.utils.tensor_dict import TensorDict
 from sample_factory.model.model_utils import get_hidden_size
 from sample_factory.utils.timing import Timing
 from sample_factory.utils.typing import Device, PolicyID
-from sample_factory.utils.utils import AttrDict, debug_log_every_n, experiment_dir, init_file_logger, log
+from sample_factory.utils.utils import AttrDict, debug_log_every_n, log
 
 
 def slice_len(s: slice) -> int:
@@ -145,8 +145,6 @@ class Batcher(StoppableEventLoopObject):
         ...
 
     def init(self):
-        init_file_logger(experiment_dir(self.cfg))
-
         device = policy_device(self.cfg, self.policy_id)
         for i in range(self.max_batches_to_accumulate):
             hidden_size = get_hidden_size(self.cfg)
--- a/sample_factory/algo/learning/batcher.py
+++ b/sample_factory/algo/learning/batcher.py
@@ -11,7 +11,7 @@ from sample_factory.algo.utils.tensor_dict import TensorDict
 from sample_factory.model.model_utils import get_hidden_size
 from sample_factory.utils.timing import Timing
 from sample_factory.utils.typing import Device, PolicyID
-from sample_factory.utils.utils import AttrDict, debug_log_every_n, log
+from sample_factory.utils.utils import AttrDict, debug_log_every_n, experiment_dir, init_file_logger, log
 
 
 def slice_len(s: slice) -> int:
@@ -145,6 +145,8 @@ class Batcher(StoppableEventLoopObject):
         ...
 
     def init(self):
+        init_file_logger(experiment_dir(self.cfg))
+
         device = policy_device(self.cfg, self.policy_id)
         for i in range(self.max_batches_to_accumulate):
             hidden_size = get_hidden_size(self.cfg)
--- a/sample_factory/algo/learning/learner_worker.py
+++ b/sample_factory/algo/learning/learner_worker.py
@@ -21,7 +21,7 @@ from sample_factory.algo.utils.torch_utils import init_torch_runtime
 from sample_factory.cfg.configurable import Configurable
 from sample_factory.utils.gpu_utils import cuda_envvars_for_policy
 from sample_factory.utils.typing import Config, PolicyID
-from sample_factory.utils.utils import log
+from sample_factory.utils.utils import experiment_dir, init_file_logger, log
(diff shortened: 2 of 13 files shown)


alex-petrenko:
> sf_log.txt will be in the same dir as the train script

this is not acceptable. Let me look at the code

alex-petrenko (changes requested):
I think I know what the problem is.

In SF1 we spawned child processes using multiprocessing.Fork which copied everything including global variables such as logger. So logger state was copied between processes.

SF2 uses multiprocessing.Spawn for a variety of reasons. As a result, it does not copy global variables.

There are two solutions:
1) Make logger a part of sf_global_context()
2) Explicitly call logger initialization in all process init method (look for a function in the beginning of every file such as learner_worker.py or inference_worker.py) 

I think I like solution #2 better. Also CFG should be available in these methods, which solves a problem of the log file being in workdir and not in traindir.

As a result we will also remove global calls to init_file_logger and configure_logger() in utils.py since they won't be needed anymore.

alex-petrenko:
Please see my review comments above! :)

sample_factory/algo/learning/batcher.py
```diff
@@ -145,6 +145,8 @@ def stop(self):
         ...
 
     def init(self):
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 148: There are way too many calls to init_file_logger.
There should be only one per process. See the process init functions in parallel_runner.

As far as I know, we only need these:
1) learner worker process
2) rollout worker process
3) inference process
4) and of course runner itself (main process)

All of the other calls are redundant and should be removed. This includes calls in init() functions

sample_factory/algo/learning/learner_worker.py
```diff
@@ -113,6 +113,8 @@ def join_batcher_thread(self):
         self.batcher_thread.join()
 
     def init(self):
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 116: Don't need this. Use init_learner process instaed

sample_factory/algo/runners/runner_parallel.py
```diff
@@ -18,6 +18,8 @@ def __init__(self, cfg):
         self.processes: List[EventLoopProcess] = []
 
     def init(self) -> StatusCode:
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 21: don't need it here. Base runner is enough

sample_factory/algo/runners/runner_serial.py
```diff
@@ -11,6 +12,8 @@ def __init__(self, cfg):
         self.trajectories_per_batch = self.cfg.batch_size // self.cfg.rollout
 
     def init(self) -> StatusCode:
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 15: Don't need this one.

sample_factory/algo/sampling/batched_sampling.py
```diff
@@ -144,6 +144,8 @@ def init(self, timing):
         Actually instantiate the env instances.
         Also creates ActorState objects that hold the state of individual actors in (potentially) multi-agent envs.
         """
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 147: Don't need this one

sample_factory/algo/sampling/inference_worker.py
```diff
@@ -141,6 +141,8 @@ def report_msg(self):
         ...
 
     def init(self, init_model_data: Optional[InitModelData]):
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 144: Don't need it here. Use init_inference_process instead

sample_factory/algo/sampling/non_batched_sampling.py
```diff
@@ -379,6 +386,7 @@ def init(self, timing: Timing):
         Actually instantiate the env instances.
         Also creates ActorState objects that hold the state of individual actors in (potentially) multi-agent envs.
         """
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 389: Don't need it here

sample_factory/algo/sampling/rollout_worker.py
```diff
@@ -125,6 +132,8 @@ def report_msg(self):
         ...
 
     def init(self):
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 135: Don't need it here

alex-petrenko: Use init_rollout_worker_process

sample_factory/algo/sampling/sampler.py
```diff
@@ -219,6 +219,8 @@ def __init__(
         self._connect_internal_components()
 
     def init(self) -> None:
+        init_file_logger(experiment_dir(self.cfg))
```
alex-petrenko, line 222: Don't need it here

sample_factory/algo/sampling/sampler.py
```diff
@@ -266,6 +268,8 @@ def __init__(
...
```
alex-petrenko, line 271: Don't need it here.

alex-petrenko (changes requested):
Please see my comments

alex-petrenko (approved):

PR #195, opened 2022-08-20
Sf2 swarm

andrewzhang505:
Integration and documentation with quad-swarm-rl.

- fixed video generation by ignoring frames that are none and making a copy of frames for when the env reuses the same object to return frames.
- Bug in _ActorCriticSeparateWeights where two params were in the wrong order compared to _ActorCriticSharedWeights 
- Copied files in swarm-rl over from the quad-swarm-rl repo to update imports from sample factory


alex-petrenko:
This is not ready to merge yet, right?

andrewzhang505:
> This is not ready to merge yet, right?

Not yet. I still need to add the gym 0.26 changes. I'll probably update this tomorrow.

docs/environment-integrations/swarm-rl.md
```diff
@@ -0,0 +1,64 @@
+# Quad-Swarm-RL Integrations
+
+### Installation
+
+Clone https://github.com/Zhehui-Huang/quad-swarm-rl into your home directory
```
alex-petrenko, line 5: Maybe mention a specific branch/commit you tested this on, because that repo might keep changing and the integration will break.

sf_examples/swarm_rl_examples/runs/multi_drone.py
```diff
@@ -0,0 +1,24 @@
...
+
+_params = ParamGrid(
+    [
+        ("seed", [0000, 1111, 2222, 3333]),
+    ]
+)
+
+SMALL_MODEL_CLI = QUAD_BASELINE_CLI + (
+    " --train_for_env_steps=10000000000 --num_workers=36 --num_envs_per_worker=4 "
+    "--quads_num_agents=8 --save_milestones_sec=10000 --async_rl=True --num_batches_to_accumulate=8 "
```
alex-petrenko, line 12: Is there a good reason to keep --num_batches_to_accumulate=8?
Can't we use the default 2?

alex-petrenko: Large value for this parameter will increase the policy lag.

andrewzhang505: I'll change it back to 2. I think this was from when I was testing the freezing issue from there being more training batches compared to learner batches but this didn't help the issue.

PR #198, opened 2022-09-27
Implements MultiInputEncoder

edbeeching:
Creates a MultiInputEncoder class that builds Encoders for any Dict observation space.
Updates dmlab and vizdoom custom models to be compatible with this change.
Includes an encoder builder test in tests/algo/test_model_builder.py

### DIFF
--- a/tests/algo/test_model_builder.py
+++ b/tests/algo/test_model_builder.py
@@ -58,8 +58,3 @@ class TestModelBuilder:
 
         assert set(encoder.obs_keys) == set(obs_space.keys())
         assert output.shape == (1, 512 * len(obs_space))
-
-
-if __name__ == "__main__":
-    t = TestModelBuilder()
-    t.test_default_encoder()
--- a/tests/algo/test_model_builder.py
+++ b/tests/algo/test_model_builder.py
@@ -1,0 +1,65 @@
+import numpy as np
+import pytest
+import torch
+from gym.spaces import Box, Dict
+
+from sample_factory.cfg.arguments import parse_sf_args
+from sample_factory.model.encoder import default_make_encoder_func
+
+
+class TestModelBuilder:
+    @pytest.mark.parametrize(
+        "obs_space",
+        [
+            Dict(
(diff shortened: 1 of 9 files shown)


tests/algo/test_model_builder.py
```diff
@@ -0,0 +1,60 @@
...
+        parser, cfg = parse_sf_args(argv=["--env=dummy"])
+        encoder = default_make_encoder_func(cfg, obs_space)
+        obs = obs_space.sample()
+        for k in obs.keys():
+            obs[k] = torch.from_numpy(np.expand_dims(obs[k], 0))
+
+        output = encoder(obs)
+
+        assert set(encoder.obs_keys) == set(obs_space.keys())
+        assert output.shape == (1, 512 * len(obs_space))
```
alex-petrenko, line 60: Love the test! <3

tests/model/test_encoder.py
```diff
@@ -0,0 +1,34 @@
...
+        obs[k] = torch.from_numpy(np.expand_dims(obs[k], 0))
+
+    output = encoder(obs)
+
+    print(encoder)
+    print(output.size())
+
+
+if __name__ == "__main__":
+    test_default_encoder()
```
alex-petrenko, line 34: What does this test do that the previous one doesn't?
Also, there are no asserts. Do you want to keep this one?

alex-petrenko (approved):
I like the changes overall. Very clean.
One thing people might want is some MLP layers between encoder and model core. But they can define a custom encoder for this I guess :)

alex-petrenko:
I don't know what's going on with the tests on this one. Looks like Github just does not want to schedule them :)
I'll merge for now since there are no merge conflicts. This will trigger automatic tests + I will test locally.

tests/model/test_encoder.py
edbeeching: Ah I think this is redundant. I was testing stuff locally. I will remove it in my next PR

PR #202, opened 2022-10-12
Adds milestone saving using periodic timers in runner

edbeeching:
Reimplements the milestone logic in the runner.


alex-petrenko (approved):
Perfect!

PR #203, opened 2022-10-13
[sf2] envpool integration

edbeeching:
Adds envpool integration for Atari and Mujoco.

The integration works as follows, when batched_sampling=True  envpool is used, when false it uses the normal method.


sf_examples/envpool_examples/atari_examples/atari/envpool_atari_params.py
```diff
@@ -0,0 +1,63 @@
+import argparse
```
alex-petrenko, line 1: Can this be maybe just envpool_examples/atari ? I feel like the directory structure is way too deep

Also, is there a chance we can reuse some of the Atari stuff between the envpool and regular Atari version?
Things like default parameters seem to be mostly the same, as well as the list of envs and things like that.

Duplicating code increases maintenance burden - the more code you have the more things to refactor each time the API changes.

sf_examples/envpool_examples/mujoco_examples/mujoco/envpool_mujoco_utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class MujocoSpec:
+    def __init__(self, name, env_id):
+        self.name = name
+        self.env_id = env_id
+
+
+MUJOCO_ENVS = [
+    MujocoSpec("mujoco_hopper", "Hopper-v4"),
```
alex-petrenko, line 24: I think this is identical to regular mujoco example too - so maybe just import from there? :)

alex-petrenko (changes requested):
Very good stuff!
There are a few comments if you're willing to simplify the code a little bit, please see the PR.

I also think we might have too many tests for this one, maybe remove a few test configurations because they're going to test the same things as regular atari/mujoco tests anyway, and tests are just going to take a lot longer :) But that's not really important, up to you.

alex-petrenko:
@edbeeching Hate to bring this up but I think Envpool caught up to Gym 0.26 as well :D
See here: https://github.com/sail-sg/envpool/releases/tag/v0.6.5

I think this is the reason why the tests fail. Your wrappers expect 4-tuple step() and just obs returned from reset() but new envpool returns 5-tuple from step() and 2-tuple from reset()

On the bright side, looks like we can delete some of the wrappers and the code will be more simple! :)
Can you merge the latest sf2 into this branch and update to envpool 0.6.5?

alex-petrenko:
@edbeeching I added a few changes to fix that Atari-only flag in enjoy while keeping the optional "smooth rendering" functionality for other envs like VizDoom. I think this should not break anything - I checked and both Atari and Doom render fine.

> sf_examples/atari_examples/atari/atari_params.py -> sf_examples/atari_examples/atari_params.py (sample for atari_utils.py)
sf_examples/atari_examples -> sf_examples/atari

I'm fine with these changes.
The reason I went with `<env>_examples` was because of isaacgymenvs having some weird relative imports where naming the module just `isaacgym` broke things.
Since this does not apply to any other envs, I think it's safe to just use <env>

alex-petrenko:
@edbeeching I made some more changes so that tests actually fail instead of freezing if there's an exception in the environment.

edbeeching:
@alex-petrenko, ah good. One less wrapper to maintain. I've updated everything.
Good point about dev_envpool.

I did the example renames as well, apart from isaacgym.  I didn't touch doom subdirs, as there are many files in each subdir.

Let me know of any other changes / comments.

alex-petrenko (approved):

PR #208, opened 2022-10-21
swarm rl documentation

andrewzhang505:
Documentation for swarm rl


alex-petrenko (approved):

alex-petrenko (changes requested):
@andrewzhang505 I checked the multi-drone video and it does not look good, the drones collide there which they should almost never do.

I think since we found a bug in the environment, we should not be reporting the buggy version. Let's remove (or comment) the multi-drone videos for now? We can add them back when Zhehui and his team fixes the bug.

Let's just keep the single-drone example for now.

PR #219, opened 2022-10-27
Lunar lander example notebook

andrewzhang505:
Added example notebook for training lunar lander and uploading to the hub. I didn't tune the hyperparameters so it doesn't train that well, but it does work. Where should I put the notebook?


alex-petrenko:
Overall this is very good, I like it!
Here's a few changes:
1) Put this into examples/notebooks
2) Notebook should not reference anything from sf_examples folder like train_gym_env, it should be self-contained because this will show how this can work with any environment, not just `Gym`.

Other than that, I think we can keep it.
As for the logging, I think you can manipulate it with logging module settings in Python, but users will probably want logging anyway? I think the way you did it now is good!

alex-petrenko (changes requested):
Please see my comment in PR for the required changes.

andrewzhang505:
It looks like there is an issue with using multiprocessing with a function defined in IPython (https://stackoverflow.com/questions/41385708/multiprocessing-example-giving-attributeerror). This is giving me an error when I try to define the `make_gym_env_func` inside the notebook instead of importing it from `sf_examples`

alex-petrenko:
Hmm if we can't use Multiprocessing in this notebook why bother using it at all... This kind of defies the whole purpose.
I guess we can still have some examples that work in serial mode, like envpool.

Okay, how about we merge this as is, but maybe you can put a short comment in the notebook explaining the situation with custom functions. Also move it to examples/notebooks.

BTW here's an interesting link about this:
https://stackoverflow.com/a/65001152/1645784

alex-petrenko (approved):

alex-petrenko:
Thank you, I think it looks good!

PR #228, opened 2022-11-12
Sf2 cfg docs

andrewzhang505:
Automatic generation of page for documenting cfg parameters. Example of how the page looks here: https://andrewzhang505.github.io/sample-factory/get-started/cfg-params/


alex-petrenko (approved):
Love this! This is a great feature!!

PR #229, opened 2022-11-14
Launcher docs fix

andrewzhang505:
Docs fix to change runner to launcher and --runner to --backend
Also fixed some incorrect module names in mujoco docs, and changed SF2 to Sample-Factory in slurm docs


alex-petrenko (approved):
<3

PR #231, opened 2022-11-26
Sf2 doom docs

andrewzhang505:
- Updated docs to include bots and multi-agent experiments
- Added wrapper for converting list of dicts to dict of lists in batched mode with multi-agents
- Some changes for generating doom videos
- Changed metric reported to HF hub to be true objective instead of reward
- Turned off record-to by default as it messes with bots envs
- Updated play doom (I couldn't figure out the controls, can only move around)


sample_factory/algo/utils/make_env.py
```diff
@@ -144,6 +158,9 @@ def __init__(self, env):
         self.is_multiagent: bool = is_multiagent
         self.num_agents: int = num_agents
 
+        if num_agents > 1:
```
alex-petrenko, line 161: I think this should be `if self.is_multiagent`

sample_factory/enjoy.py
```diff
@@ -183,11 +185,11 @@ def max_frames_reached(frames):
                         finished_episode[agent_i] = True
                         rew = episode_reward[agent_i].item()
                         episode_rewards[agent_i].append(rew)
-                        if cfg.use_record_episode_statistics:
-                            if "episode" in infos[agent_i].keys():
-                                reward_list.append(infos[agent_i]["episode"]["r"])
-                        else:
-                            reward_list.append(rew)
+                        # if cfg.use_record_episode_statistics:
```
alex-petrenko, line 188: Why did you comment this out?
I think we need this for Atari (@wmFrank knows the details)

sf_examples/vizdoom/doom/doom_params.py
```diff
@@ -41,9 +41,10 @@ def add_doom_env_eval_args(parser):
     """Arguments used only during evaluation."""
     parser.add_argument(
         "--record_to",
-        default=join(os.getcwd(), "..", "recs"),
+        # default=join(os.getcwd(), "..", "recs"),
+        default=None,
         type=str,
-        help="Record episodes to this folder.",
+        help="Record episodes to this folder. It is recommended to use --save_video instead of --record_to for bots environments.",
```
alex-petrenko, line 47: Those are very different things.

--record_to records a demo that can be replayed in high resolution, even FullHD.

save_video can only render at resolution of the env observations.

Can you change this comment so it explains this difference?

sf_examples/vizdoom/doom/multiplayer/doom_multiagent_wrapper.py
```diff
@@ -217,13 +218,22 @@ def __init__(self, num_agents, make_env_func, env_config, skip_frames, render_mo
...
 
-    def set_reward_shaping(self, reward_shaping: dict, agent_idx: int):
-        self.current_reward_shaping[agent_idx] = reward_shaping
-        self.set_env_attr(
-            agent_idx,
-            "unwrapped.reward_shaping_interface.reward_shaping_scheme",
-            reward_shaping,
-        )
+    def set_reward_shaping(self, reward_shaping: dict, agent_indices: Union[int, slice]):
+        if isinstance(agent_indices, slice):
```
alex-petrenko, line 222: You're repeating the same code here twice 
If `agent_indices` is not a slice just convert it to a slice of lenght 1 and use the same for-loop code! :)

sf_examples/vizdoom/doom/multiplayer/doom_multiagent_wrapper.py
```diff
@@ -351,20 +361,26 @@ def render(self):
...
+        elif self.render_mode == "rgb_array":
+            obs_display = [o["obs"] for o in self.last_obs]
+            obs_grid = concat_grid(obs_display, self.render_mode)
+            return obs_grid
         else:
-            obs_display = self.last_obs[0]["obs"]
-            cv2.imshow("vizdoom", cvt_doom_obs(obs_display))
+            raise ValueError(f"{self.render_mode=} is not supported")
+
+        # render_multiagent = True
```
alex-petrenko, line 374: Do we need the commented out code?

alex-petrenko (changes requested):
Really like your work! The rendering code looks great and the self-play videos are awesome!

I requested a few small changes, please take a look!

alex-petrenko:
Maybe also add a small test for your new wrapper so that the coverage does not drop!

tests/examples/test_example_multi.py
```diff
@@ -27,14 +27,16 @@ def run_test_env_multi(cfg: Config, eval_cfg: Config, **kwargs):
...
+    def test_sanity(self, async_rl: bool, train_steps: int, batched_sampling: bool):
         cfg, eval_cfg = default_multi_cfg()
         cfg.async_rl = async_rl
         cfg.train_for_env_steps = train_steps
-        cfg.num_workers = 1
-        cfg.batch_size = 128
+        cfg.num_workers = 2
+        cfg.batch_size = 256
         cfg.serial_mode = True
+        cfg.batched_sampling = batched_sampling
```
alex-petrenko, line 39: Why did you add this?

tests/examples/test_example_multi.py
```diff
@@ -49,6 +51,7 @@ def test_example_multi(self):
         cfg.train_for_env_steps = 100000
         cfg.num_workers = 8
         cfg.batch_size = 512
+        cfg.batched_sampling = True
```
alex-petrenko, line 54: Why did you change this to batched sampling?

I think for multi-agent we actually want to test non-batched version because some of the features (multi-policy training and inactive agents are actually only supported in non-batched mode)

sf_examples/vizdoom/doom/doom_params.py
```diff
@@ -41,9 +41,10 @@ def add_doom_env_eval_args(parser):
     """Arguments used only during evaluation."""
     parser.add_argument(
         "--record_to",
-        default=join(os.getcwd(), "..", "recs"),
+        # default=join(os.getcwd(), "..", "recs"),
+        default=None,
         type=str,
-        help="Record episodes to this folder.",
+        help="Record episodes to this folder. This records a demo that can be replayed at full resolution. Currently, this does not work for bot environments so it is recommended to use --save_video to record episodes at lower resolution instead of such environments",
```
alex-petrenko, line 47: Misprint, I think should be `instead FOR such environments`)

alex-petrenko (changes requested):

alex-petrenko:
Couple of small things, please take a look.

Also, after you fix these please also merge latest `sf2` into this branch, let's see if new Ming's solution fixes the tests.

tests/examples/test_example_multi.py
andrewzhang505: I added this to test the new wrapper since it also returns the obs as a list

tests/examples/test_example_multi.py
andrewzhang505: Sure, I'll change back to non-batched

tests/examples/test_example_multi.py
andrewzhang505: from the coverage, it seems the new wrapper is not fully covered since the `BatchedDictObservationsWrapper` already converts to dict. I can look into adding more tests to fully cover the wrapper

PR #232, opened 2022-11-28
Sf2 ci fix

wmFrank:
added precheck tests to the ci workflow

### DIFF
--- a/.github/workflows/test-ci.yml
+++ b/.github/workflows/test-ci.yml
@@ -5,6 +5,7 @@ on:
     branches:
     - master
     - sf2
+    - sf2-ci-fix
   pull_request:
     branches:
     - master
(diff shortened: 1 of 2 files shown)


wmFrank:
I found the errors are all related to torch memory access. The official doc says it could be problems if the machine has low limits for shared memory. And I printed every layer output size for the failed tests and cannot find the output size in the error message. Also found a link here(https://discourse.julialang.org/t/is-it-normal-that-github-actions-hang/73653) which also has a hanging problem. Currently, I added a pre-check tests to the ci workflow to do a pre-check of the running environment.

wmFrank:
Do multi-policies tests or pbt tests take up much more memory than single test? I noticed that nearly all the failed tests are multi-policies tests or pbt tests. And it could reach the github action limits sometimes.

PR #234, opened 2022-11-30
isaac gym docs w/o videos

andrewzhang505:


sample_factory/utils/utils.py
```diff
@@ -487,3 +487,7 @@ def log_every_n(n, _level, msg, *args, **kwargs):
 
 def debug_log_every_n(n, msg, *args, **kwargs):
     log_every_n(n, logging.DEBUG, msg, *args, **kwargs)
+
+
+def get_top_level_script():
+    return argv[0].split("sample-factory/")[-1][:-3].replace("/", ".")
```
alex-petrenko, line 493: I think this is maybe okay for now, but it will fail if the train script is not in Sample Factory codebase (like with quadrotors for example)

One idea is to get the module path directly from the module object (you can do module.__spec__.name)

alex-petrenko (approved):
Great feature, thank you!!

PR #238, opened 2022-11-30
adds additional isaacgym envs

edbeeching:
Can you check the hyperparams vs. https://github.com/NVIDIA-Omniverse/IsaacGymEnvs/tree/main/isaacgymenvs/cfg/train


alex-petrenko (approved):
Thank you! 
I think the parameters look good. 
We should train some policies with these envs at least as a sanity check.

PR #250, opened 2022-12-14
Fixed bug when using mixed action spaces

edbeeching:
When using tuple action distributions in non-batched mode, there is a sneaky bug that appears. This is due to np.split and torch.split not having the same functionality. See: https://github.com/pytorch/pytorch/issues/50012

Fixed by using the cumsum of the action splits when in non batched mode,

### DIFF
--- a/sample_factory/algo/learning/learner.py
+++ b/sample_factory/algo/learning/learner.py
@@ -758,7 +758,7 @@ class Learner(Configurable):
 
                     kl_old_mean = float(kl_old.mean().item())
                     recent_kls.append(kl_old_mean)
-                    if kl_old.max().item() > 100:
+                    if kl_old.numel() > 0 and kl_old.max().item() > 100:
                         log.warning(f"KL-divergence is very high: {kl_old.max().item():.4f}")
 
                 # update the weights
--- a/sample_factory/algo/sampling/non_batched_sampling.py
+++ b/sample_factory/algo/sampling/non_batched_sampling.py
@@ -148,7 +148,9 @@ class ActorState:
             return self._process_action_space(actions, is_discrete=False)
         elif isinstance(self.env_info.action_space, gym.spaces.Tuple):
             out_actions = []
-            for split, space in zip(np.split(actions, self.env_info.action_splits), self.env_info.action_space):
+            for split, space in zip(
+                np.split(actions, np.cumsum(self.env_info.action_splits)), self.env_info.action_space
+            ):
                 is_discrete = isinstance(space, gym.spaces.Discrete)
                 out_actions.append(self._process_action_space(split, is_discrete))
             return out_actions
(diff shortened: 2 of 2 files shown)


sample_factory/algo/sampling/non_batched_sampling.py
```diff
@@ -148,7 +148,9 @@ def curr_actions(self) -> np.ndarray | List | Any:
             return self._process_action_space(actions, is_discrete=False)
         elif isinstance(self.env_info.action_space, gym.spaces.Tuple):
             out_actions = []
-            for split, space in zip(np.split(actions, self.env_info.action_splits), self.env_info.action_space):
+            for split, space in zip(
+                np.split(actions, np.cumsum(self.env_info.action_splits)), self.env_info.action_space
+            ):
```
alex-petrenko, line 153: I checked and it looks like on my machine all three versions give different results:

```
>>> import numpy as np
>>> import torch
>>> actions = [1,2,3,4,5,6]
>>> actions_t = torch.tensor(actions)
>>> 
>>> splits = [2,4]
>>> 
>>> np.split(actions, splits)
[array([1, 2]), array([3, 4]), array([5, 6])]
>>> np.split(actions, np.cumsum(splits))
[array([1, 2]), array([3, 4, 5, 6]), array([], dtype=int64)]
>>> torch.split(actions_t, splits)
(tensor([1, 2]), tensor([3, 4, 5, 6]))
```

I think it would be nice to avoid that discrepancy.

alex-petrenko: Here's one possible solution:

```
>>> splits_relative = splits
>>> splits_absolute = np.cumsum(splits_relative)[:-1]
>>> 
>>> np.split(actions, splits_absolute)
[array([1, 2]), array([3, 4, 5, 6])]
```

alex-petrenko: It will also work even if splits is only len=1, so should be safe!

alex-petrenko (changes requested):

sample_factory/algo/sampling/non_batched_sampling.py
edbeeching: That empty array is so strange. Good spot, thanks. I have updated the PR.

edbeeching:
I bundled a fix for #251 in to this PR as well

sample_factory/algo/learning/learner.py
```diff
@@ -758,7 +758,7 @@ def _train(
 
                     kl_old_mean = float(kl_old.mean().item())
                     recent_kls.append(kl_old_mean)
-                    if kl_old.max().item() > 100:
+                    if kl_old.numel() > 0 and kl_old.max().item() > 100:
```
alex-petrenko, line 761: This will fix the crash but I still don't understand how we got a 0-dim tensor here...

alex-petrenko (approved):

PR #254, opened 2022-12-31
Added Multi GPU

Mikahil:
Added cfg entry "gpu_per_policy" which defaults to 1 and creates batchers, samplers, and learners to match the value. When its greater than one, the learners will use DistributedDataParallel to synchronize the policy's parameters.


alex-petrenko:
@Mikahil if not too much trouble, consider running `make format` next time before committing. This will make sure pre-commit checks pass and you can actually run the tests on GH actions.

https://www.samplefactory.dev/12-community/contribution/#how-to-create-a-pull-request

PR #256, opened 2023-01-11
Added script name params for hf model card

andrewzhang505:
Fixed the training and enjoy script names in the HF model card


alex-petrenko (approved):
Wow fantastic!! Thank you :)

Issue #261, opened 2023-01-15
ValueError in colab


amrzv:
Hi.
When running `samplefactory_hub_example.ipynb` notebook in colab the `ValueError` raises:
![image](https://user-images.githubusercontent.com/36787333/212562627-24829a22-84fe-468b-92d0-0a65bdebc223.png)

Seems that there is a bug in `sample_factory.utils.gpu_utils.py`.

alex-petrenko:
looks like `CUDA_VISIBLE_DEVICES` contains something like `0,\n` and we don't expect this.

alex-petrenko:
Pushed a small fix, see if this helps :)

https://github.com/alex-petrenko/sample-factory/commit/7509cc4b37cbf5b7b541ea0dda49cbad1152c8fe

amrzv:
I forgot to mention, that this error was raised when running in colab without GPU. Now, when running in colab without GPU the `ValueError` above is gone, but next another raises:
<img width="668" alt="image" src="https://user-images.githubusercontent.com/36787333/212642718-f3ef3d62-508c-4a1a-922f-588a2e95860b.png">
Seems that the device should be manually set before in `cfg`:
`cfg.device = 'gpu' if torch.cuda.is_available() else 'cpu'`

However, when I add this line to `cfg`, restart notebook and run again I still observe device=='gpu' in logging information:
<img width="769" alt="image" src="https://user-images.githubusercontent.com/36787333/212643890-6d2e34a1-9648-427f-9fe5-a82d9602c74f.png">

I'm not sure if everything is supposed to work on CPU.

When running in colab with GPU whole notebook works without errors.

andrewzhang505:
From the screenshot, it looks like the notebook is loading an old experiment where the cfg still has the device set to gpu. Could you try changing the `experiment_name` and running again with `cfg.device=cpu`? Alternatively, you can add the command line `--restart_behavior=restart` to the argv to wipe out the old experiment

amrzv:
I cleared the local dir, restarted the kernel and run again. Now notebook works without errors:
![image](https://user-images.githubusercontent.com/36787333/212678131-43c42d33-b900-4b8d-88c9-06f0feb4b5f3.png)

So, probably this line should be added as on the screenshot:
`cfg.device = 'gpu' if torch.cuda.is_available() else 'cpu'`

alex-petrenko:
@andrewzhang505 thank you for fixing this!

[...]

PR #262, opened 2023-01-16
Set device in notebook

andrewzhang505:
Set device to cpu in notebook if cuda not available for https://github.com/alex-petrenko/sample-factory/issues/261


alex-petrenko (approved):
Thank you!!

PR #272, opened 2023-06-26
[DOC] Fix incorrect call to register_model_core_factory

paLeziart:
Hello,

This PR fixes a small typo in the documentation about the call to the `def register_model_core_factory(self, make_model_core_func: MakeCoreFunc):` function of the `ModelFactory` class.

It only affects a single line of a documentation file.


alex-petrenko (approved):
Thank you!!

PR #278, opened 2023-07-17
Checking for vtrace before adding advantage stats to summary

alex404:
So running simulations where `with_vtrace=True` currently crashes due to trying to load stats about the advantage function. Adding this check fixes it.

I'm not sure where this bug got introduced, because I've regularly used vtrace in the past. Also just FYI, in my simulations, I find that setting `with_vtrace=True` and `normalize_returns=False` yields better performance for feedforward networks, but RNNs more or less require GAE and normalized returns.

### DIFF
--- a/sample_factory/algo/learning/learner.py
+++ b/sample_factory/algo/learning/learner.py
@@ -866,10 +866,13 @@ class Learner(Configurable):
         stats.act_min = var.mb.actions.min()
         stats.act_max = var.mb.actions.max()
 
-        stats.adv_min = var.mb.advantages.min()
-        stats.adv_max = var.mb.advantages.max()
-        stats.adv_std = var.adv_std
-        stats.adv_mean = var.adv_mean
+        if not self.cfg.with_vtrace:
+
+            stats.adv_min = var.mb.advantages.min()
+            stats.adv_max = var.mb.advantages.max()
+            stats.adv_std = var.adv_std
+            stats.adv_mean = var.adv_mean
+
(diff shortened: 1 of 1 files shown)


alex-petrenko:
@alex404 thank you for the fix! Can I please ask you to run the code formatting before we merge such that the pre-commit check passes?
The instruction is here https://www.samplefactory.dev/12-community/contribution/#how-to-create-a-pull-request
You basically just need to run `make format` in the code folder, and before that you should run `pip install .[dev]` to install `black, isort`, etc.

sample_factory/algo/learning/learner.py
```diff
@@ -866,10 +866,13 @@ def _record_summaries(self, train_loop_vars) -> AttrDict:
         stats.act_min = var.mb.actions.min()
         stats.act_max = var.mb.actions.max()
 
-        stats.adv_min = var.mb.advantages.min()
-        stats.adv_max = var.mb.advantages.max()
-        stats.adv_std = var.adv_std
-        stats.adv_mean = var.adv_mean
+        if not self.cfg.with_vtrace:
```
alex-petrenko, line 869: Instead of checking for vtrace I suggest we check something like `"adv_mean" in stats` because there can be other reasons advantage stats are not present. Good catch though!!

alex-petrenko (approved):

sample_factory/algo/learning/learner.py
alex404: Yah of course, sorry for not noticing the documentation. When I find a moment I'll update the PR.

PR #289, opened 2023-12-29
Add nethack to sf_examples

BartekCupial:
## Overview

This PR introduces NetHack to Sample Factory. 
The code is based on three repositories:
- https://github.com/Miffyli/nle-sample-factory-baseline 
- https://github.com/dungeonsdatasubmission/dungeonsdata-neurips2022
- https://github.com/BartekCupial/sample-factory

#### Easy installation and experiments
By adding NetHack to examples in Sample Factory my main goal is to improve reproducibility and allow for easy experimentation with NetHack as I found that there were many issues with installation and experiments. For example when trying to reproduce experiments in D&D repository Dockerfile required fixing moolib library and Cmake.txt file.  Additionally since D&D repo implemented APPO from scratch (implementation details in RL matter) I found that just my moving to SF I've managed to increase the score from 2k to about 2.8k.

#### Additional metrics
Sample Factory supports logging of additional policy stats if any are found in `info["episode_extra_stats"]`. I've added wrappers which log blstats and additional auxiliary scores. Look at `sf_examples/nethack/utils/task_rewards.py]`.

#### `render_mode=rgb_array`
NLE natively doesn't support rgb_array. I've added rgb_array mode for rendering by using  `RenderCharImagesWithNumpyWrapperV2` wrapper introduced by https://github.com/Miffyli/nle-sample-factory-baseline.
By using `rgb_array` in enjoy we can save and examine episodes. Example: https://github.com/BartekCupial/sample-factory/assets/92169405/47884b73-beeb-4303-a72f-75d202aa87a8

#### Evaluation
NetHack can have very long episodes (100k env steps) and additionally since the environment is highly stochastic we usually need a lot of evaluation episodes to measure the policy performance (usually 1024). I highly recommend using recently introduced `eval.py` since using enjoy would be really long.


docs/09-environment-integrations/nethack.md
```diff
@@ -0,0 +1,69 @@
...
+git clone https://github.com/facebookresearch/nle.git nle && cd nle \
+&& git checkout v0.9.0 && git submodule init && git submodule update --recursive \
+&& sed '/#define NLE_ALLOW_SEEDING 1/i#define NLE_ALLOW_SEEDING 1' include/nleobs.h -i \
+&& sed '/self\.nethack\.set_initial_seeds = f/d' nle/env/tasks.py -i \
+&& sed '/self\.nethack\.set_current_seeds = f/d' nle/env/tasks.py -i \
+&& sed '/self\.nethack\.get_current_seeds = f/d' nle/env/tasks.py -i \
+&& sed '/def seed(self, core=None, disp=None, reseed=True):/d' nle/env/tasks.py -i \
+&& sed '/raise RuntimeError("NetHackChallenge doesn.t allow seed changes")/d' nle/env/tasks.py -i \
+&& sed -i '/def render(self, mode="human"):/a\        if not self.last_observation:\n            return' nle/env/base.py \
+&& python setup.py install && cd ..
```
klyuchnikova-ana, line 21: Had issues with this in Python 3.11 conda env (pybind11 errors), but worked fine in 3.10.

Can we add a note that 3.10 is recommended and there are known issues in 3.11?

docs/09-environment-integrations/nethack.md
```diff
@@ -0,0 +1,69 @@
...
+&& sed '/self\.nethack\.get_current_seeds = f/d' nle/env/tasks.py -i \
+&& sed '/def seed(self, core=None, disp=None, reseed=True):/d' nle/env/tasks.py -i \
+&& sed '/raise RuntimeError("NetHackChallenge doesn.t allow seed changes")/d' nle/env/tasks.py -i \
+&& sed -i '/def render(self, mode="human"):/a\        if not self.last_observation:\n            return' nle/env/base.py \
+&& python setup.py install && cd .. 
+
+# install sample factory with nethack extras
+pip install -e .[nethack]
+pip install -e sf_examples/nethack/render_utils
+```
```
klyuchnikova-ana, line 27: Had to additionally install `conda install -c conda-forge pybind11` to make this work

I wonder if it's also a good idea to rename this package into something more specific to avoid any potential name conflicts?

Like `nethack-render` or `nethack-render-utils`

I'm referring to the name of the python package/module. (also, I don't feel particularly strongly about this, we can keep it as is)

docs/09-environment-integrations/nethack.md
```diff
@@ -0,0 +1,69 @@
...
+To train a model in the `nethack_challenge` environment:
+
+```
+python -m sf_examples.nethack.train_nethack --env=nethack_challenge --batch_size=4096 --num_workers=16 --num_envs_per_worker=30 --character=mon-hum-neu-mal --experiment=nethack_monk
+```
+
+To visualize the training results, use the `enjoy_nethack` script:
+
+```
+python -m sf_examples.nethack.enjoy_nethack --env=nethack_challenge --character=mon-hum-neu-mal --experiment=nethack_monk
```
klyuchnikova-ana, line 43: This worked! (I think)

setup.py
```diff
@@ -14,6 +14,16 @@
...
 _mujoco_deps = ["gymnasium[mujoco]", "mujoco<2.5"]
+_nethack_deps = [
+    "numba ~= 0.58",
+    "pandas ~= 2.1",
+    "matplotlib ~= 3.8",
+    "seaborn ~= 0.12",
+    "scipy ~= 1.11",
+    "shimmy",
+    "tqdm ~= 4.66",
+    "debugpy ~= 1.6",
```
klyuchnikova-ana, line 25: Today I learned!
Didn't know about this `~=` feature

sf_examples/nethack/README.md
```diff
@@ -0,0 +1,53 @@
+## Installation
+
+To install NetHack, you need nle and its dependencies.
+
+```bash
+# nle dependencies
+apt-get install build-essential python3-dev python3-pip python3-numpy autoconf libtool pkg-config libbz2-dev
+conda install cmake flex bison lit
+
```
klyuchnikova-ana, line 9: My preference would be to avoid duplicating the README as you'd need to update & maintain both files.
Can we have a very brief readme here that just points at documentation?

klyuchnikova-ana (approved):
Great stuff!

A few minor comments, but overall looks super clean! :fire:

docs/09-environment-integrations/nethack.md
BartekCupial: I'm fine with renaming it. Done.

sf_examples/nethack/README.md
BartekCupial: That's true, I've added the documentation later and forgot to remove it. Done.

docs/09-environment-integrations/nethack.md
BartekCupial: Actually in setup.py we have `python_requires=">=3.8",`, but at the same time we use features from python 3.10 (typehints `Tensor | float`). Maybe we should bump python_requires to 3.10.

BartekCupial: Added some info in docs about issues with python 3.11 and above.

setup.py
BartekCupial: This is actually a useful feature. It allows bug fix releases but avoids introducing new features that could potentially break your code.

docs/09-environment-integrations/nethack.md
BartekCupial: :)

docs/09-environment-integrations/nethack.md
klyuchnikova-ana: > Actually in setup.py we have `python_requires=">=3.8",`, but at the same time we use features from python 3.10 (typehints `Tensor | float`). Maybe we should bump python_requires to 3.10.

We run unit tests on 3.8, 3.9, and 3.10 and all currently pass.
`Tensor | float` work because there's some import "from future import annotations" or smth like that

BartekCupial: I guess I forgot to use that import in some parts of my other code. Thanks :)

klyuchnikova-ana (approved):
👍

PR #298, opened 2024-04-29
Fix for Windows training

Ivan-267:
I was able to start Sample Factory training on Windows 10 64 bit using the following: 

- First installing the PR branch for signal slot that allows installation on Windows: https://github.com/alex-petrenko/signal-slot/pull/3,
- Installing Sample Factory using `pip install sample-factory` (then applying the modifications from this PR which changes the UNIX specific PWD module, which caused an error for me on Windows, to one that should hopefully work on both OS types),
- Device set to `CPU` (Otherwise there some CUDA detection errors, but I do not have a CUDA device on this PC, so I didn't experiment with that further),
- Serial mode set to `True`,

I have tested this only with my custom env that uses Godot RL Agents, but I assume it should work for all supported envs.

It needs to be checked if it still works fine on Linux after these changes, I hope the automated tests on the repository will cover this case.


alex-petrenko (changes requested):
This requires adding `getpass` to dependencies right?

Can we instead use try/catch on `pwd.getpwuid` and if it fails use `import getpass; return getpass...`?

alex-petrenko:
Alternatively, if `getpass` is easily available on all platforms, maybe we can just add it to the list of requirements in setup.py?

Ivan-267:
> This requires adding `getpass` to dependencies right?
> 
> Can we instead use try/catch on `pwd.getpwuid` and if it fails use `import getpass; return getpass...`?

I think `getpass` comes with Python and shouldn't need pip install or etc. I didn't use it before however, so I can check more later.

alex-petrenko (approved):

alex-petrenko:
Thank you!

PR #305, opened 2024-09-15
Support PettingZoo Parallel API and action mask

nkzawa:
As the title, add support for PettingZoo by providing a wrapper class. Also, add support for action masks.

### Creating PettingZoo env

```python
from sample_factory.envs.pettingzoo_envs import PettingZooParallelEnv

def make_pettingzoo_env(full_env_name, cfg=None, env_config=None, render_mode: Optional[str] = None):
    return PettingZooParallelEnv(some_env.parallel_env(render_mode=render_mode))
```

Currently, it supports only [Parallel API](https://pettingzoo.farama.org/api/parallel/) since it requires a different flow of executions for supporting [AEC API](https://pettingzoo.farama.org/api/aec/). 

### Action mask

It works when you add the `action_mask` key to the `dict` observation.

```python
import gymnasium as gym

class CustomEnv(gym.Env):
    def __init__(self, full_env_name, cfg, render_mode: Optional[str] = None):
        self.observation_space = gym.spaces.Dict({
            "obs": gym.spaces.Box(low=0, high=1, shape=(3, 3, 2), dtype=np.int8),
            "action_mask": gym.spaces.Box(low=0, high=1, shape=(9,), dtype=np.int8),
        })
        self.action_spaces = gym.spaces.Discrete(9)

    def step(self, action):
        ...
        return {"obs": obs, "action_mask": action_mask}, reward, terminated, truncated, info
```

It seems this is the most common interface for providing action masks on PettingZoo, so I think it makes sense to follow that.  It's also common to have the value in `info` as `info["action_mask"]` but it's not supported for now since sample-factory requires knowing the shape to allocate buffer as far as I understand.

I added an example that trains Tic-Tac-Toe but I'm not sure about the configuration so appreciate any suggestions.

-------------

btw, I followed `CONTRIBUTING.md` but `make check-codestyle` fails with the error:

```sh
.../python3.11/site-packages/sympy/polys/numberfields/resolvent_lookup.py: "pyflakes[F]" failed during execution due to RecursionError('maximum recursion depth exceeded')
```

Also, `make test` fails when `mujoco` is involved.

```sh
OSError: dlopen(/System/Library/OpenGL.framework/OpenGL, 0x0006): tried: '/System/Library/OpenGL.framework/OpenGL' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/System/Library/OpenGL.framework/OpenGL' (no such file), '/System/Library/OpenGL.framework/OpenGL' (no such file, not in dyld cache)
```

Python version is 3.11.9 on Intel Mac

### DIFF
--- a/docs/07-advanced-topics/action-masking.md
+++ b/docs/07-advanced-topics/action-masking.md
@@ -1,6 +1,6 @@
 # Action Masking
 
-Action masking is a technique used to restrict the set of actions available to an agent in certain states. This can be particularly useful in environments where some actions are invalid or undesirable in specific situations. Sample Factory supports action masking, allowing you to implement this feature in your custom environments.
+Action masking is a technique used to restrict the set of actions available to an agent in certain states. This can be particularly useful in environments where some actions are invalid or undesirable in specific situations. See [paper](https://arxiv.org/abs/2006.14171) for more details.
 
 ## Implementing Action Masking
 
--- a/docs/07-advanced-topics/action-masking.md
+++ b/docs/07-advanced-topics/action-masking.md
@@ -1,0 +1,38 @@
+# Action Masking
+
+Action masking is a technique used to restrict the set of actions available to an agent in certain states. This can be particularly useful in environments where some actions are invalid or undesirable in specific situations. Sample Factory supports action masking, allowing you to implement this feature in your custom environments.
+
+## Implementing Action Masking
+
+To implement action masking in your environment, you need to add an `action_mask` field to the observation dictionary returned by your environment. Here's how to do it:
+
+1. Define the action mask space in your environment's observation space
+2. Generate and return the action mask in both `reset()` and `step()` methods
+
+Here's an example of a custom environment implementing action masking:
+
+```python
+import gymnasium as gym
+import numpy as np
+
+class CustomEnv(gym.Env):
+    def __init__(self, full_env_name, cfg, render_mode=None):
+        ...
+        self.observation_space = gym.spaces.Dict({
+            "obs": gym.spaces.Box(low=0, high=1, shape=(3, 3, 2), dtype=np.int8),
+            "action_mask": gym.spaces.Box(low=0, high=1, shape=(9,), dtype=np.int8),
+        })
+        self.action_space = gym.spaces.Discrete(9)
+
+    def reset(self, **kwargs):
+        ...
+        # Initial action mask that allows all actions
+        action_mask = [1, 1, 1, 1, 1, 1, 1, 1, 1]
+        return {"obs": obs, "action_mask": action_mask}, info
+
+    def step(self, action):
+        ...
+        # Generate new action mask based on the current state
+        action_mask = [1, 0, 0, 1, 1, 1, 0, 1, 1]
+        return {"obs": obs, "action_mask": action_mask}, reward, terminated, truncated, info
+```
--- a/docs/07-advanced-topics/action-masking.md
+++ b/docs/07-advanced-topics/action-masking.md
@@ -27,12 +27,12 @@ class CustomEnv(gym.Env):
     def reset(self, **kwargs):
         ...
         # Initial action mask that allows all actions
-        action_mask = [1, 1, 1, 1, 1, 1, 1, 1, 1]
+        action_mask = np.array([1, 1, 1, 1, 1, 1, 1, 1, 1])
         return {"obs": obs, "action_mask": action_mask}, info
 
     def step(self, action):
         ...
         # Generate new action mask based on the current state
-        action_mask = [1, 0, 0, 1, 1, 1, 0, 1, 1]
+        action_mask = np.array([1, 0, 0, 1, 1, 1, 0, 1, 1])
         return {"obs": obs, "action_mask": action_mask}, reward, terminated, truncated, info
 ```
--- a/docs/09-environment-integrations/pettingzoo.md
+++ b/docs/09-environment-integrations/pettingzoo.md
@@ -1,0 +1,46 @@
+# PettingZoo
+
+[PettingZoo](https://pettingzoo.farama.org/) is a Python library for conducting research in multi-agent reinforcement learning. This guide explains how to use PettingZoo environments with Sample Factory.
+
+## Installation
+
+Install Sample Factory with PettingZoo dependencies with PyPI:
+
+```bash
+pip install -e sample-factory[pettingzoo]
+```
+
+## Running Experiments
+
+Run PettingZoo experiments with the scripts in `sf_examples`.
+The default parameters are not tuned for throughput.
+
+To train a model in the `tictactoe_v3` environment:
+
+```
+python -m sf_examples.train_pettingzoo_env --algo=APPO --env=tictactoe_v3 --experiment="Experiment Name"
+```
+
+To visualize the training results, use the `enjoy_pettingzoo_env` script:
+
+```
+python -m sf_examples.enjoy_pettingzoo_env --env=tictactoe_v3 --experiment="Experiment Name"
+```
+
+Currently, the scripts in `sf_examples` are set up for the `tictactoe_v3` environment. To use other PettingZoo environments, you'll need to modify the scripts or add your own as explained below.
+
+### Adding a new PettingZoo environment
+
+To add a new PettingZoo environment, follow the instructions from [Custom environments](../03-customization/custom-environments.md), with the additional step of wrapping your PettingZoo environment with `sample_factory.envs.pettingzoo_envs.PettingZooParallelEnv`.
+
+Here's an example of how to create a factory function for a PettingZoo environment:
+
+```python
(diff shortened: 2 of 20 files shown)


alex-petrenko:
hi @nkzawa !
thank you for this contribution

would it be possible to make sure pre-commit passes, as well as Ubuntu tests?
https://www.samplefactory.dev/12-community/contribution/ there are some details here

A short doc page would be very nice to have too :) I will do a proper review soon!

tests/envs/pettingzoo/test_pettingzoo.py
```diff
@@ -0,0 +1,64 @@
...
+from tests.envs.utils import eval_env_performance
+from tests.utils import clean_test_dir
+
+
+class TestPettingZooEnv:
+    @pytest.fixture(scope="class", autouse=True)
+    def register_pettingzoo_fixture(self):
+        register_custom_components()
+        yield  # this is where the actual test happens
+        reset_global_context()
```
alex-petrenko, line 20: Added this (similar to other env tests)
this resets global encoder factory, otherwise other tests can fail if this runs first.

(I'd be the first to admit this is not a perfect solution but hey I wrote this years ago)

alex-petrenko (changes requested):
Great contribution, thank you!
There are a couple of small comments to address.
If you have the time to add a small documentation page, that'd be very helpful! Something similar to other contributed env integrations, like https://www.samplefactory.dev/09-environment-integrations/nethack/ (does not have to be as elaborate as this, installation instruction and training example would already be great, a wandb run/report or some videos - doubly great!)

alex-petrenko:
Ubuntu tests seem to pass, macOS tests have this random spurious failure but seems unrelated to the PR. Other than the small comments above, I'm happy to merge this

sample_factory/algo/utils/action_distributions.py
```diff
@@ -104,12 +110,22 @@ def log_probs(self):
...
-        samples = torch.multinomial(self.probs, 1, True)
+    def sample(self, action_mask=None):
+        probs = self.probs
+        if action_mask is not None:
+            probs = probs * action_mask
+            all_zero = (probs.sum(dim=-1) == 0).unsqueeze(-1)
+            epsilons = torch.full_like(probs, 1e-6)
+            probs = torch.where(all_zero, epsilons, probs)  # ensure sum of probabilities is non-zero
+
+        samples = torch.multinomial(probs, 1, True)
```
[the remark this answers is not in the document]
nkzawa: Technically, it seems there is no need to add up to 1, according to the doc:

> The rows of input do not need to sum to one (in which case we use > the values as weights), ...

https://pytorch.org/docs/stable/generated/torch.multinomial.html

But I'm not so sure honestly (I'm a newbie on RL). So please feel free to fix if you see something wrong with the code.

sample_factory/model/actor_critic.py
```diff
@@ -321,4 +336,13 @@ def create_actor_critic(cfg: Config, obs_space: ObsSpace, action_space: ActionSp
     from sample_factory.algo.utils.context import global_model_factory
 
     make_actor_critic_func = global_model_factory().make_actor_critic_func
-    return make_actor_critic_func(cfg, obs_space, action_space)
+    return make_actor_critic_func(cfg, obs_space_without_action_mask(obs_space), action_space)
```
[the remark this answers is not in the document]
nkzawa: I assumed that it meant to do inside `default_make_actor_critic_func`. Fixed it so anyway 🙏

nkzawa:
Will add docs as well 👍

sample_factory/algo/utils/action_distributions.py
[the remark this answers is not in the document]
nkzawa: Actually, seems it requires to normalize the value with softmax as far as I understand, so implemented that.

nkzawa:
Based on [this paper](https://arxiv.org/abs/2006.14171) and the [Maskable PPO implementation in Stable-Baselines3](https://sb3-contrib.readthedocs.io/en/master/modules/ppo_mask.html), it appears that action masks should also be applied when calculating log probabilities as well. This approach helps the model learn to avoid selecting invalid actions, as far as I understand. I'll modify the code accordingly.

docs/07-advanced-topics/action-masking.md
```diff
@@ -0,0 +1,38 @@
...
+
+```python
+import gymnasium as gym
+import numpy as np
+
+class CustomEnv(gym.Env):
+    def __init__(self, full_env_name, cfg, render_mode=None):
+        ...
+        self.observation_space = gym.spaces.Dict({
+            "obs": gym.spaces.Box(low=0, high=1, shape=(3, 3, 2), dtype=np.int8),
```
alex-petrenko, line 22: Small nit: I wonder if low=0 high=1 here is intentional, would this mean binary observations?

I understand 0/1 in action_mask since this is a binary mask

docs/07-advanced-topics/action-masking.md
```diff
@@ -0,0 +1,38 @@
+# Action Masking
```
alex-petrenko, line 1: This documentation is wonderful, thank you!

docs/09-environment-integrations/pettingzoo.md
```diff
@@ -0,0 +1,46 @@
+# PettingZoo
```
alex-petrenko, line 1: Love this. Thank you!

sample_factory/algo/utils/action_distributions.py
```diff
@@ -81,35 +81,71 @@ def argmax_actions(distribution):
         raise NotImplementedError(f"Action distribution type {type(distribution)} does not support argmax!")
 
 
+# Retrieved from AllenNLP:
+# https://github.com/allenai/allennlp/blob/80fb6061e568cb9d6ab5d45b661e86eb61b92c82/allennlp/nn/util.py#L243
+def masked_softmax(logits, mask):
+    # To limit numerical errors from large vector elements outside the mask, we zero these out.
+    result = functional.softmax(logits * mask, dim=-1)
```
alex-petrenko, line 88: Can you help me understand this please?

I think logits in general can be negative, or positive but close to 0, in which case multiplying them by zero does not achieve the desired effect.

I'd say we should probably use something like this instead?

```
def masked_softmax(logits, mask):
    # Mask out the invalid logits by adding a large negative number (-1e9)
    logits = logits + (mask == 0) * -1e9
    result = functional.softmax(logits, dim=-1)
    result = result * mask
    result = result / (result.sum(dim=-1, keepdim=True) + 1e-13)
    return result
```

the choice of 1e-9 is arbitrary here, but it could be something like `-max(abs(logits)) * 1e6` to make this universal

sample_factory/algo/utils/action_distributions.py
```diff
@@ -81,35 +81,71 @@ def argmax_actions(distribution):
...
+    return result
+
+
+# Retrieved from AllenNLP:
+# https://github.com/allenai/allennlp/blob/80fb6061e568cb9d6ab5d45b661e86eb61b92c82/allennlp/nn/util.py#L286
+def masked_log_softmax(logits, mask):
+    # vector + mask.log() is an easy way to zero out masked elements in logspace, but it
+    # results in nans when the whole vector is masked.  We need a very small value instead of a
+    # zero in the mask for these cases.
+    logits = logits + (mask + 1e-13).log()
```
alex-petrenko, line 100: This makes more sense to me, this is essentially adding log(1e-13) to non-valid elements which is about -30. I'm not sure if this is universally correct, but most likely should work. Why can't we just explicitly add a large negative constant though, like -1e9 or `-max(abs(logits)) * 1e6` like in the previous example?

docs/07-advanced-topics/action-masking.md
nkzawa: This is intentional since it's retrieved from [tic tac toe of PettingZoo](https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/classic/tictactoe/tictactoe.py#L135) but I think we can change if it's confusing.

sample_factory/algo/utils/action_distributions.py
nkzawa: It's got from AllenNLP including the comment so don't fully understand but as far as I investigated seems your version is safer in some cases even tho usually the results are identical in both versions 👍

sample_factory/algo/utils/action_distributions.py
nkzawa: It seems you're correct. This version causes a problem in extreme cases as far as I tested.

sample_factory/algo/utils/action_distributions.py
nkzawa: fixed

sample_factory/algo/utils/action_distributions.py
nkzawa: fixed

alex-petrenko (approved):

PR #306, opened 2024-10-21
Add ONNX Model Export Support

nkzawa:
This PR introduces support for exporting trained models to the ONNX format in Sample Factory. The feature will be useful in many cases, including for people who want to use their trained models on a different platform or language. It's nice to build the functionality into Sample Factory, since usually, the underlying architecture is something that users don't need to know about.

It's implemented using [TorchScript-based ONNX Exporter](https://pytorch.org/docs/stable/onnx_torchscript.html) for now. I tried [TorchDynamo](https://pytorch.org/docs/stable/onnx_dynamo.html) as well but couldn’t get it to work. 
The usage is described in the added documentation.

### DIFF
--- a/.github/workflows/cover-ci.yml
+++ b/.github/workflows/cover-ci.yml
@@ -30,7 +30,7 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install conda env & dependencies
       run: |
-       pip install -e '.[atari, mujoco, envpool]'
+       pip install -e '.[atari, mujoco, envpool, onnx]'
        conda list
     - name: Install codecov dependencies
       run: |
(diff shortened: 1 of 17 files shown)


alex-petrenko:
Thank you for pushing these!
Looks like there are some merge conflicts and some syntax errors with Python 3.8.
Lmk if you can fix it, otherwise we can probably drop python 3.8 support at this point

nkzawa:
Conflict and syntax errors on python3.8 should be fixed. Also, support for the action mask was added in the ONNX export.

nkzawa:
@alex-petrenko I think it's ready to get review

alex-petrenko (approved):
This looks amazing! 🔥

PR #307, opened 2024-10-29
Log env_steps in wandb

BartekCupial:
This PR adds metric definitions for Weights & Biases logging to better organize and visualize training metrics. Now we have an option to use `env_steps` in X axis.

### DIFF
--- a/sample_factory/utils/wandb_utils.py
+++ b/sample_factory/utils/wandb_utils.py
@@ -55,6 +55,14 @@ def init_wandb(cfg):
 
     wandb.config.update(cfg, allow_val_change=True)
 
+    wandb.define_metric("train/env_steps")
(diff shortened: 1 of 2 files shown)


sample_factory/utils/wandb_utils.py
```diff
@@ -55,6 +55,14 @@ def init_wandb_func():
 
     wandb.config.update(cfg, allow_val_change=True)
 
+    wandb.define_metric("train/env_steps")
```
alex-petrenko, line 58: This can probably be a loop over a list of patterns to make the code a bit cleaner

alex-petrenko (approved):
👍 thanks!

@BartekCupial can you explain what exactly this does? I believe by default in wandb the "global_step" is the number of env steps, so will this just add the same metric under a different name (env_steps)? What does `define_metric` do in this case?

PR #308, opened 2024-11-18
Feature: add wandb dir as parser argument

fabioseel:
Adds the possibility to change the wandb directory.
Default behaviour (when not setting it) should be the same as before.


sample_factory/cfg/cfg.py
```diff
@@ -744,6 +744,12 @@ def add_wandb_args(p: ArgumentParser):
         nargs="*",
         help="Tags can help with finding experiments in WandB web console",
     )
+    p.add_argument(
```
alex-petrenko, line 747: 👍

alex-petrenko (approved):

PR #330, opened 2026-01-13
train_for_seconds flag not being tracked

lmBored:
This is a small PR, previously `train_for_seconds` flag was not being tracked (as `total_train_seconds` was not calculated), and as a result training didn't stop as set by `train_for_seconds`.

I believe this was a minor overlook, thus I didn't open an issue but open a PR directly.

### DIFF
--- a/sample_factory/algo/runners/runner.py
+++ b/sample_factory/algo/runners/runner.py
@@ -678,6 +678,7 @@ class Runner(EventLoopObject, Configurable):
         self._observers_call(AlgoObserver.on_connect_components, self)
 
     def _should_end_training(self):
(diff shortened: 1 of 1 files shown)


alex-petrenko:
Hi @lmBored ! Sorry for the delay
What's the reason for closing this? Seems like a legit fix :)

lmBored:
Hi! Thank you for your response! It was such a small fix, so I thought it would be better to integrate with a larger PR so you don't need to approve twice:)

But Im glad you are here! I will reopen the PR

lmBored:
oh I forgot this branch is still a PR

PR #331, opened 2026-01-29
feat: add DQN/IDQN

lmBored:
In my project, Sample Factory was used in editable mode as a standalone folder, that's why all changes appear in a single commit in this PR (I merged the whole folder instead of cherry-picking)

Also pls note that some minor reformatting in `cfg.py` is included. These changes were applied automatically by my formatter, but they improve consistency with the surrounding code, so I kept them.

### Implementation:
+ Double DQN
+ Prioritized Experience Replay
+ Independent q_head for multi-agent
+ For async setting, I calculate the number of updates need to be made by subtracting full update debt (num_updates * steps_per_update), then cap the actual work accordingly
+ Disable sample actions

### Performance (Tested on Doom environments - single agent):
+ `dqn_max_updates_per_batch=1`: 12.5% faster than APPO
+ `dqn_max_updates_per_batch=4`: Same speed as APPO
+ ~Converges slower than APPO~

### Limitations:
+ Requires tuning
+ Performance generally is bottlenecked by `dqn_max_updates_per_batch`

### Example train command:
```bash
python -m sf_examples.vizdoom.train_vizdoom \
    --env=doom_battle \
    --algo=DQN \
    --use_rnn=False \
    --train_for_env_steps=5000
```

Let me know what you think about this approach!

### DIFF
--- a/sample_factory/algo/learning/learner_dqn.py
+++ b/sample_factory/algo/learning/learner_dqn.py
@@ -1,0 +1,442 @@
+from __future__ import annotations
+
+import copy
+from typing import Dict, Optional, Union
+
+import torch
+from torch import Tensor
+from torch.nn import functional as F
+
+from sample_factory.algo.learning.learner import Learner
+from sample_factory.algo.utils.env_info import EnvInfo
+from sample_factory.algo.utils.misc import LEARNER_ENV_STEPS, POLICY_ID_KEY, STATS_KEY, TRAIN_STATS, memory_stats
+from sample_factory.algo.utils.model_sharing import ParameterServer
+from sample_factory.algo.utils.replay_buffer import PrioritizedReplayBuffer, ReplayBuffer
+from sample_factory.algo.utils.rl_utils import prepare_and_normalize_obs
+from sample_factory.algo.utils.tensor_dict import TensorDict, shallow_recursive_copy
+from sample_factory.algo.utils.torch_utils import synchronize, to_scalar
+from sample_factory.utils.attr_dict import AttrDict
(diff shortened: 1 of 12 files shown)


alex-petrenko:
Hi @lmBored ! The PR looks really good!

I think at this point I won't be able to commit to maintaining and testing more code (as you can see by the failing tests, I barely have cycles to take care of the existing codebase :) )

I highly encourage you to start a fork and use this PR and similar issues to advertise your implementation! I have some people on discord who were asking about offline methods.

I wonder what your results were like with DQN on Doom? I remember all these years ago I got much better results with A2C/PPO compared to DQN and it was one of the reasons for focusing purely on online algorithms here. Do you have any training curves or behavior videos? 🎮

lmBored:
Hi @alex-petrenko, thank you for your response and looking at the PR. I understand your point, I've also just realised how old sample factory is now :)

Fitting an off-policy method to an asynchronous framework is quite troublesome, but being able to implement offline methods or model-based methods in the same fashion as sample factory would be rlly interesting. Making a fork is a good idea though, I'll consider that after I'm done with my current project.

This is a run (in a scenario where agent runs forward while jumping and avoiding lava pits) with [PPO](https://wandb.ai/comrad/marl_vizdoom/runs/doom_pitfall_APPO_20260128_234008_20260128_234008_251908), and this is with [DQN](https://wandb.ai/comrad/marl_vizdoom/runs/doom_pitfall_DQN_20260211_110619_20260211_110619_563045). Here FPS of DQN is worse than PPO is because I'm doing 4 updates per batch. At 2 updates, it performs on par with PPO, and at 1 update, DQN performs better than PPO.

PR #286, opened 2023-10-06
Fix vtrace

alex404:
Sorry for taking ages to get to this... got distracted by ICLR.

Anyway, I ran make format, and more closely followed the instructions in your docs.

If I need to make any further changes just let me know, and I'll get to it faster this time around.

PR #332, opened 2026-06-30
Fix advantage summaries never being logged in _record_summaries

danielpalenicek:
PR https://github.com/alex-petrenko/sample-factory/pull/286 guarded the advantage-stat logging with `if "adv_mean" in stats:` to avoid the V-trace
  AttributeError on `var.mb.advantages` (which exists only on the GAE path). But
  `adv_mean` is only ever added to `stats` *inside* that block, so the condition
  is always False: the fix is dead code and adv_min/max/std/mean are now never
  logged under any config.
  
  Guard on the value that's actually conditional instead:
  
      if "advantages" in var.mb:
          stats.adv_min = var.mb.advantages.min()
          stats.adv_max = var.mb.advantages.max()
          stats.adv_std = var.adv_std
          stats.adv_mean = var.adv_mean

  GAE -> logged; V-trace -> skipped, no crash. Verified on CartPole both paths.
  (`if "adv_mean" in var:` would be wrong too — it's always True and re-crashes.)

### DIFF
--- a/sample_factory/algo/learning/learner.py
+++ b/sample_factory/algo/learning/learner.py
@@ -870,7 +870,7 @@ class Learner(Configurable):
         stats.act_min = var.mb.actions.min()
         stats.act_max = var.mb.actions.max()
 
-        if "adv_mean" in stats:
+        if "advantages" in var.mb:
             stats.adv_min = var.mb.advantages.min()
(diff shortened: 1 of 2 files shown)


alex-petrenko:
Thanks!

